PR #77, opened 2021-11-11
Update pull request template

mbaker3:
### PR
- Remove auto notification of team that no longer exists
- Improve structure of template

I'm open to any feedback on the template. I don't think this feels like too much to fill out for a PR while also forcing the submitter to think through the changes they've made during submission.

### Notify

@scratch-games/anvil-pr-notifiers


jkeon (approved):
I like it

PR #78, opened 2022-03-24
UpdateHandle - Add IsSourceUpdating

mbaker3:
Add boolean flags to the update sources and `UpdateHandle` to indicate when they are within their update phase. This allows executing code to take shortcuts if it knows it's currently executing during an appropriate time in the frame.

**Example:** A system needs to conduct work during `LateUpdate` but doesn't know if it's executing during that phase (or during the `UpdateHandle`'s update call stack) it can check this flag to execute its work immediately instead of scheduling to wait for the next `LateUpdate`.

The feature does not exist

`UpdateHandle.IsSourceUpdating` is exposed for consumers.

None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/DelayedExecution/UpdateSources/AbstractUpdateSource.cs
+++ b/DelayedExecution/UpdateSources/AbstractUpdateSource.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Diagnostics;
 using Anvil.CSharp.Core;
 
 namespace Anvil.CSharp.DelayedExecution
@@ -13,6 +14,11 @@ namespace Anvil.CSharp.DelayedExecution
         /// </summary>
         public event Action OnUpdate;
 
+        /// <summary>
+        /// Indicates whether the source is currently executing its OnUpdate phase.
+        /// </summary>
+        public bool IsUpdating { get; private set; }
+
         protected AbstractUpdateSource()
         {
         }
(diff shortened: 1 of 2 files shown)


DelayedExecution/UpdateSources/AbstractUpdateSource.cs
```diff
@@ -21,12 +27,19 @@ protected override void DisposeSelf()
         {
             UpdateHandleSourcesManager.RemoveUpdateSource(this);
             OnUpdate = null;
+
             base.DisposeSelf();
         }
 
         protected void DispatchOnUpdateEvent()
         {
+            Debug.Assert(!IsUpdating);
```
jkeon, line 36: Are these `Debug.Assert`'s necessary? 

Seems heavy handed for an update loop and might be better off as a special more intense debugging?

jkeon (approved):
Looks good, just one question.

DelayedExecution/UpdateSources/AbstractUpdateSource.cs
mbaker3: It's a bit extra cautious but with so many variations on update source implementations it catches what would be a difficult to debug situation.

This performance cost is only in debug builds. If we find it's really slowing things down we can put it under a deeper debugging condition but, really, this is 2 boolean checks * a handful of update sources. So maybe an additional 10 boolean checks?

PR #80, opened 2022-04-27
Feature - Log Listener

mbaker3:
This PR introduces three changes.
 - Add `Log.IsHandlingLog` to communicate when `Log` is currently in the phase of iterating through log handlers.
 - Add `Log.SuppressLogging` to allow developers to suppress log output to handlers. This is not for release builds, it's used to suppress logs that the developer doesn't want to emit but can't stop the output of. (Ex: false flag warnings)
 - ILogListener - Add ability to redirect logging from other systems into `Log` for its handlers to process.

#### Tag Along
 - Fixed spelling mistakes in doc comments
 - Better organize files in Anvil.Csharp.Logging.csproj

 - There is no indication to developers when `Log` is handling a log message.
 - There is no way to suppress log messages from being handled
 - There is not a standardized way to redirect another logging system into logger. Of course, this could be accomplished before but wasn't explicitly supported.
    - This work supports (https://github.com/decline-cookies/anvil-unity-core/pull/58)

#### `Log.IsHandlingLog`
 - Returns true while a log is being evaluated by handlers. Returns false at all other times.
#### `Log.SuppressLogging`
 - Defaults to false. While set to true handling of any incoming log messages is skipped.

#### ILogListener
This implementation, through reflection, automatically discovers all implementations of `ILogListener` with the `DefaultLogListenerAttribute` and instantiates them. 

Aside from instantiate there is no interaction or management of any listeners by the `Log` class. The `ILogListener` interface is currently just for easy discovery of implementations and to allow for easy maintenance/changes if `Log` to `ILogListener` interaction is required.

A `ILogListener` implementation, when instantiated is expected to handle hooking itself into the logging system it's designed to listen too and calling the most appropriate `Logger` methods. Implementations must also take care to not introduce recursion if there is also an `ILogHandler` implementation for the system it's listening too. `Log.IsHandlingLog` can be a helpful part of the solution to this.

 - Capturing logs for source that you cannot change or don't want to convert to use `Logger` instances.
 - Allowing projects to get the benefits of `Log` without the effort required to convert all existing log calls to use `Logger` instances.

 - None

 - [ ] Yes
 - [x] No


mbaker3:
Once the premise and implementation is approved I'll do a documentation pass

mbaker3:
Still open to feedback but merging and can revise in a new PR.

PR #82, opened 2022-05-10
Extension Methods - Remove namespaces add docs

mbaker3:
Remove name spaces from Extension classes and add top level documentation where missing.
Extension classes shouldn't have namespaces so that they are easier for developers to discover (source: #2). Anvil didn't consistently apply this rule until now.

Some extension method classes have namespaces that require the developer to import before using.

All extension method classes exist at the root namespace.

Developer discover ability of namespaces

 - None

 - [ ] Yes 
 - [x] No - You may have unused import statements code now.

### DIFF
--- a/Collections/ListExtension.cs
+++ b/Collections/ListExtension.cs
@@ -1,5 +1,8 @@
 ﻿using System.Collections.Generic;
 
+/// <summary>
+/// Extension methods for use with an <see cref="IList{T}"/> instance.
+/// </summary>
 public static class ListExtension
 {
     /// <summary>
--- a/Reflection/TypeExtension.cs
+++ b/Reflection/TypeExtension.cs
@@ -1,34 +1,34 @@
-﻿namespace System.Reflection
+﻿
+using System;
+
+/// <summary>
+/// A set of convenience extensions for <see cref="Type"/>
+/// </summary>
+public static class TypeExtension
 {
     /// <summary>
-    /// A set of convenience extensions for <see cref="Type"/>
+    /// Resolves the default value of a provided runtime type.
+    /// The runtime equivalent of `default(MyType)`
     /// </summary>
-    public static class TypeExtension
+    /// <param name="type">The type to resolve the default value for</param>
+    /// <returns>The default value for the given type</returns>
+    /// <remarks>Stolen from: https://stackoverflow.com/a/2490274 </remarks>
+    public static object CreateDefaultValue(this Type type)
     {
-        /// <summary>
-        /// Resolves the default value of a provided runtime type.
-        /// The runtime equivalent of `default(MyType)`
-        /// </summary>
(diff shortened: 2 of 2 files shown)


tjvezina:
Instead of adding to the namespace that the extension is intended for, could you use your own for all extension classes? `namespace Anvil.CSharp.Extensions`?

Just out of curiosity, why does putting them in a namespace hinder discoverability? Do the extensions just not appear in autocomplete?

mbaker3:
> Instead of adding to the namespace that the extension is intended for, could you use your own for all extension classes? `namespace Anvil.CSharp.Extensions`?
> 
> Just out of curiosity, why does putting them in a namespace hinder discoverability? Do the extensions just not appear in autocomplete?

Some IDE configurations don't show intellisense for options in namespaces that haven't been imported into the current context. Others require a different key combination than the standard one used to bring up intellisense.

While it's pretty normal to press the extra key combo when writing a type it's less common to think to do so when writing the method portion of a statement. Having extension methods outside of a namespace allows the method to always appear when working with an instance of the target type.

tjvezina:
Are name conflicts not a concern then? I know Anvil only has a couple of users today, but even if you tried to include some third-party library one day that defines its own `TypeExtension` class with no namespace, you'd have compile errors until someone changes the name right?

tjvezina:
I would vote for a single `.Extensions` namespace for all static extension classes, easier to maintain and include. Speaking personally, I tend to have a template file for C# or Unity projects, and would just include said extensions namespace by default.

This isn't worth changing at this point, but out of curiosity, was there a reason to use `Anvil.CSharp.<Area>` over `Anvil.<Area>` in namespaces and similar?

mbaker3:
Alright, you guys make good arguments. I went and tested VSCode and Visual Studio for Mac. Either by default or after changing some preferences both behave the same way as rider with regard to discoverability of extension methods in an un-imported namespace. 

My mind was stuck in the past, we live in the future now and IDEs are better.

With that in mind we should just put the extension methods in the same namespace as their siblings in the same directory.

Will close this PR and adjust the approach.

> This isn't worth changing at this point, but out of curiosity, was there a reason to use Anvil.CSharp.<Area> over Anvil.<Area> in namespaces and similar?

It's been a while but I think the rationale was that this was the Pure C#/.NET variant of Anvil. C#/.NET was the platform while `anvil-unity-core` is `Anvil.Unity.*`

Maybe we should have instead built onto the `Anvil.CSharp.` prefix in the Unity module to make it `Anvil.CSharp.Unity...`
 That way you could say
What library is this? -> Anvil
What language is it implemented in? -> Csharp
What is the specific platform that's required -> Unity.

That way in the future, let's say Boo in Unity took off again. You could have an `Anvil.Boo.Unity` that would be separate from Anvil.Csharp.Unity`.

...but we didn't and, like you said, I'm not sure it's worth changing at this point.

PR #83, opened 2022-05-17
Hotfix extensions

mbaker3:
Revisions in response to discussion on #82 

The diffs are a little insane so the following is a summary of changes:
 - `ListExtension` -> Moved to namespace `Anvil.CSharp.Collections`
 - `LINQExtension` -> Moved to namespace `Anvil.CSharp.Linq`
 - `TypeExtension` -> Moved to namespace `Anvil.CSharp.Reflection`
 - `FileLogHandler.cs` -> Moved to a `LogHandler/FIleLogHandler.cs`

Extension classes have inconsistent namespaces

Extension classes have namespaces that reflect their location in the library. Consistent with all other types.

namespace inconsistency

 - None

 - [x] Yes - You'll need to import the appropriate namespaces if using the affected extensions
 - [ ] No


Collections/ListExtension.cs
```diff
@@ -1,21 +1,27 @@
...
+        /// <summary>
+        /// Uniquely adds a collection of items to an <see cref="IList{T}"/>.
+        /// Useful for when LINQ can't be used or the underlying list instance must remain the same.
+        /// </summary>
+        /// <param name="list">The list to add to.</param>
+        /// <param name="items">The items to attempt to add.</param>
+        public static void AddUnique<T>(this IList<T> list, IEnumerable<T> items)
         {
-            if (!list.Contains(item))
+            foreach (var item in items)
```
tjvezina, line 18: So now `var` is acceptable? 🤨

tjvezina (approved):

Collections/ListExtension.cs
mbaker3: hey now, that's not a real change in this PR...
Fixed anyway. Shamefully that was my original commit!

PR #86, opened 2022-06-10
Add CollectionExtension with DisposeAllAndClear

mbaker3:
Add `CollectionExtension` with `DisposeAllAndClear` method.

Developers have to build their own iterations to dispose of a collection of `IDisposable` elements.
THEN they have to clear the collection manually. Can you believe it?!

With one easy extension method call developers can now have their collection's elements disposed and then cleared out.

None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Collections/CollectionExtension.cs
+++ b/Collections/CollectionExtension.cs
@@ -9,9 +9,9 @@ namespace Anvil.CSharp.Collections
     public static class CollectionExtension
     {
         /// <summary>
-        /// Dispose all elements of a <see cref="ICollection{T}"/> then clear.
+        /// Dispose all elements of a <see cref="ICollection{T}"/> and then clear the collection.
         /// </summary>
-        /// <param name="collection">A collection of elements to dispose then clear.</param>
+        /// <param name="collection">The <see cref="ICollection{T}"/> to operate on.</param>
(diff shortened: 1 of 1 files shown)


Collections/CollectionExtension.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Anvil.CSharp.Collections
+{
+    /// <summary>
+    /// Extension methods for use with <see cref="ICollection{T}"/> instances.
+    /// </summary>
+    public static class CollectionExtension
+    {
+        /// <summary>
+        /// Dispose all elements of a <see cref="ICollection{T}"/> then clear.
```
jkeon, line 12:
```suggestion
        /// Dispose all elements of a <see cref="ICollection{T}"/> and then clear the collection.
```

Collections/CollectionExtension.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// </summary>
+        /// <param name="collection">A collection of elements to dispose then clear.</param>
```
jkeon, line 14:
```suggestion
        /// <param name="collection">The <see cref="ICollection{T}"/> to operate on.</param>
```

jkeon (approved):
Looks good, minor tweaks to docs

Issue #87, opened 2022-06-18
Bring `UnbufferedStreamReader` to feature parity with `StreamReader`


mbaker3:
`UnbufferedStreamReader`'s features have been developed on an as needed basis. Any additional features should be logged against this task.

At some point we should define the remaining features that `UnbufferedStreamReader` is missing.

PR #88, opened 2022-06-18
Unbuffered Stream Reader

mbaker3:
Adds `UnbufferedStreamReader`. A stream reader that does not buffer data from the backing BaseStream.
This implementation is less performant than `StreamReader` but doesn't read ahead on the underlying stream ensuring the BaseStream's position to remains lockstep with the characters read out of `UnbufferedStreamReader`.

This stream is useful when reading files with mixed encoding or where the BaseStream needs to be advanced the minimum amount required.

`StreamReader` fills a buffer of characters when responding to read requests. The characters buffer is filled by repeatedly reading blocks of its `BaseStream` and converting them into characters until the character buffer is full.

This means that the `BaseStream` is advanced beyond the point that the `StreamReader` has been read.

`UnbufferedStreamReader` effectively functions the same as `StreamReader` with a byte and character buffer size set to 1. The `UnbufferedStreamReader` reads one byte at a time out of `BaseStream` until it has enough to resolve a full character. That one character is then evaluated by the read method to decide whether to continue reading or return.

**Note:** The implementation does not currently cover the complete features set of `StreamReader`. New features will be developed as needed. This effort is captured in #87 

#### Why don't you just calculate actual the `BaseStream` position based on `StreamReader`'s unused buffer contents?
TL;DR
 - Invalid bytes are replaced with a pre-defined fallback character which may not equal the byte size of the invalid bytes.
 - The reflection relies heavily on Microsoft not changing the way `StreamReader` is implemented
 - It does not work for variable-byte encodings (Ex: UTF-7)

I went down that path. It's not possible. [This StackOverflow post](https://stackoverflow.com/questions/5404267/c-sharp-streamreader-and-seeking/17457085#17457085) covers the issue in detail. Read the linked post + all comments to understand.

If you really want to dig into this read through `StreamReader`'s source: https://referencesource.microsoft.com/#mscorlib/system/io/streamreader.cs,b5fe1efcec14de32

None, This class allows developers to continue reading a `BaseStream` beyond the lifecycle of the `UnbufferedStreamReader`.

 - None

 - [ ] Yes
 - [x] No

### DIFF
--- a/IO/UnbufferedStreamReader.cs
+++ b/IO/UnbufferedStreamReader.cs
@@ -31,7 +31,7 @@ namespace Anvil.CSharp.IO
 
         // Used for APIs that require a buffer. All single length;
         private readonly byte[] m_SingleByteBuffer = new byte[1];
-        private readonly char[] m_charBuffer = new char[1];
+        private readonly char[] m_CharBuffer = new char[1];
 
         private bool m_IsPreambleHandled;
 
@@ -67,7 +67,7 @@ namespace Anvil.CSharp.IO
             finally
             {
                 m_StringBuilder.Clear();
-                
+
                 base.DisposeSelf();
             }
         }
@@ -106,7 +106,7 @@ namespace Anvil.CSharp.IO
 
                 if (character == default || character == '\r' || character == '\n')
                 {
-                    return RenderAndClearStringBuilder();
+                    return ResolveAndClearStringBuilder();
                 }
 
                 m_StringBuilder.Append(character);
@@ -130,7 +130,7 @@ namespace Anvil.CSharp.IO
 
                 if (character == default)
                 {
-                    return RenderAndClearStringBuilder();
+                    return ResolveAndClearStringBuilder();
                 }
 
                 m_StringBuilder.Append(character);
@@ -171,13 +171,13 @@ namespace Anvil.CSharp.IO
                     return default;
                 }
 
-                decodedCharCount = m_Decoder.GetChars(m_SingleByteBuffer, 0, 1, m_charBuffer, 0);
+                decodedCharCount = m_Decoder.GetChars(m_SingleByteBuffer, 0, 1, m_CharBuffer, 0);
             } while (decodedCharCount < 1);
 
-            return m_charBuffer[0];
+            return m_CharBuffer[0];
         }
 
-        private string RenderAndClearStringBuilder()
+        private string ResolveAndClearStringBuilder()
         {
             string result = m_StringBuilder.ToString();
             m_StringBuilder.Clear();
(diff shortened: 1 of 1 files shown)


IO/UnbufferedStreamReader.cs
```diff
@@ -0,0 +1,187 @@
...
+namespace Anvil.CSharp.IO
+{
+    /// <summary>
+    /// A stream reader that does not use a buffer.
+    /// <see cref="StreamReader"/> is more performant but advances its <see cref="BaseStream"/> past its own position an indeterminate amount.
+    /// This stream is useful when reading files with mixed encoding or where the <see cref="BaseStream"/> needs to be advanced the minimum amount required.
+    /// </summary>
+    /// <remarks>
+    /// This reader is not currently at feature parity with <see cref="StreamReader"/>. Features will be developed as needed.
+    /// See: #87
```
jkeon, line 15: Should we link that SO post in here too? Useful background info.

IO/UnbufferedStreamReader.cs
```diff
@@ -0,0 +1,187 @@
...
+        /// </summary>
+        public Encoding CurrentEncoding { get; }
+
+        private readonly bool m_LeaveBaseStreamOpen;
+        private readonly Decoder m_Decoder;
+        private readonly StringBuilder m_StringBuilder;
+
+        // Used for APIs that require a buffer. All single length;
+        private readonly byte[] m_SingleByteBuffer = new byte[1];
+        private readonly char[] m_charBuffer = new char[1];
```
jkeon, line 34:
```suggestion
        private readonly char[] m_CharBuffer = new char[1];
```

IO/UnbufferedStreamReader.cs
```diff
@@ -0,0 +1,187 @@
...
+                    return default;
+                }
+
+                decodedCharCount = m_Decoder.GetChars(m_SingleByteBuffer, 0, 1, m_charBuffer, 0);
+            } while (decodedCharCount < 1);
+
+            return m_charBuffer[0];
+        }
+
+        private string RenderAndClearStringBuilder()
```
jkeon, line 180: Seems odd to call it `Render`?  Maybe `ResolveAndClearStringBuilder`? `ExtractAndClear`?

jkeon (changes requested):
Minor comments, otherwise looks good

IO/UnbufferedStreamReader.cs
mbaker3: I mean, you are rendering a string from the builder's component parts but sure I've rename it to resolve.

mbaker3:
Feedback addressed

IO/UnbufferedStreamReader.cs
jkeon: Not to the screen man.... not to the screen.

jkeon (approved):

PR #91, opened 2022-08-02
Implement `OneTimeLogger` + Minor Refactor + `ILogger`

mbaker3:
Implements the ability to configure a log message to only ever emit one time, per session, from a given call site. This allows developers to log information from a location that gets executed frequently (frame updates) without spamming the console.

Move `Log.Logger` out into its own type rather than remaining nested under the `Log` type. Also fixed some spelling mistakes.

Create ILogger to help keep all Logger implementation APIs in sync

**Tag Along**
 - Configured directories in the Logger project to exclude sub-directories from namespace contribution. This allows Rider to suggest the correct namespaces for types.

 - There is no way to one time log
 - The `Logger` type is declared within the `Log` type.
 - Various `Logger` implementation APIs are kept in sync manually.

#### One Time Logging
The developer can now declare a one time log by: `myLogger.OneTime().Warning("This log will only be emitted once per session");`
The `OneTimeLogger` generates a hash for each call site based on the caller path and line number. This hash is stored in a hash set on first call and used to check whether the call has been issued before.

`OneTimeLogger.Reset()` can be called to reset the log tracking and allow all one time messages to be emitted once more.

#### Move `Logger` out of `Log`
Moved out since there are now multiple implementations of `Logger` it didn't make sense to keep `Logger` as an inner type of `Log`.

Re Spelling: "Suppress" was incorrectly spelled throughout the types 🤦 

#### ILogger
All (sort of) Logger implementations now implement the `ILogger` interface. This will help us keep the Logger APIs in sync.
 - The one Logger that can't implement `ILogger` is [BurstableLogger](https://github.com/decline-cookies/anvil-unity-dots/blob/main/Scripts/Runtime/Logging/BurstableLogger.cs). This is fine, BurstableLogger is a very restricted logger that barely works. As the Burst compiler gains features there's a good chance we'll be able to bring that logger inline with `ILogger`.

 - None

 - None

 - [x] Yes
 - [ ] No

- All references to `Log.Logger` must be changed to `Logger`.
- Any use of `Log.SupressLogging` must be changed to `Log.SuppressLogging`

### DIFF
--- a/Logging/.CSProject/Logger/OneTimeLogger.cs
+++ b/Logging/.CSProject/Logger/OneTimeLogger.cs
@@ -1,0 +1,86 @@
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+
+namespace Anvil.CSharp.Logging
+{
+    /// <summary>
+    /// A wrapper for <see cref="Logger"/> that only allows a message to be emitted from a given call site once per session.
+    /// A call site is determined to be unique by a combination of its file path and line number.
+    /// </summary>
+    public readonly struct OneTimeLogger : ILogger
+    {
+        private static readonly HashSet<int> s_CalledLogSites = new HashSet<int>();
+
+        /// <summary>
+        /// Resets <see cref="OneTimeLogger"/>'s to allow one time logs to be emitted one more time.
+        /// </summary>
+        public static void Reset()
+        {
+            s_CalledLogSites.Clear();
+        }
+
+
+        private readonly Logger m_Logger;
+
+
+        internal OneTimeLogger(in Logger logger)
+        {
+            m_Logger = logger;
+        }
+
+        /// <inheritdoc cref="ILogger.Debug"/>
+        public void Debug(object message, [CallerFilePath] string callerPath = "", [CallerMemberName] string callerName = "", [CallerLineNumber] int callerLine = 0)
+        {
+            if (!ShouldEmitLog(callerPath, callerLine))
+            {
+                return;
+            }
+
+            m_Logger.Debug(message, callerPath, callerName, callerLine);
+        }
+
+        /// <inheritdoc cref="ILogger.Warning"/>
+        public void Warning(object message, [CallerFilePath] string callerPath = "", [CallerMemberName] string callerName = "", [CallerLineNumber] int callerLine = 0)
+        {
+            if (!ShouldEmitLog(callerPath, callerLine))
+            {
+                return;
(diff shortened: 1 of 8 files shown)


Logging/.CSProject/Logger/OneTimeLogger.cs
```diff
@@ -0,0 +1,86 @@
...
+    {
+        private static readonly HashSet<int> s_CalledLogSites = new HashSet<int>();
+
+        /// <summary>
+        /// Resets <see cref="OneTimeLogger"/>'s to allow one time logs to be emitted one more time.
+        /// </summary>
+        public static void Reset()
+        {
+            s_CalledLogSites.Clear();
+        }
```
tjvezina, line 20: Is there a use case for this? While I'm not sure how it could be done, allowing a single log to appear again seems useful, resetting all logs across the project less so

Logging/.CSProject/Logger/OneTimeLogger.cs
```diff
@@ -0,0 +1,86 @@
...
+
+
+        private readonly Logger m_Logger;
+
+
+        internal OneTimeLogger(in Logger logger)
```
tjvezina, line 26: Are the double newlines intentional?

tjvezina (approved):

Logging/.CSProject/Logger/OneTimeLogger.cs
tjvezina: Alright, I've read the Unity PR now :P I should've guessed that this covers an editor-only problem, and isn't likely to be actually used in a normal context.

Logging/.CSProject/Logger/OneTimeLogger.cs
mbaker3: Yea, I like to double new line between inner class blocks, static blocks, field/property blocks, and method blocks.

If it really bugs someone we can have a little style fight 🤷 . It's not a big deal in Rider but I find it helps readability in editors that don't insert annotations above declaration lines.

Logging/.CSProject/Logger/OneTimeLogger.cs
mbaker3: 😄 Yea. Unity domain reload specific feature.

I could see it potentially used in a situation where an application soft resets (post update?) but even then, depending on how the application domains are managed the newly loaded code would likely have its own version of logger.

PR #94, opened 2022-08-02
Add DictionaryExtension - Dispose Values Convenience

mbaker3:
Add `DictionaryExtension.DisposeAllValuesAndClear`.

 - None

`DictionaryExtension.DisposeAllValuesAndClear` will dispose all values in a `Dictionary` and then clear the `Dictionary`

 - None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Collections/DictionaryExtension.cs
+++ b/Collections/DictionaryExtension.cs
@@ -1,0 +1,24 @@
+using System;
+using System.Collections.Generic;
+
+namespace Anvil.CSharp.Collections
+{
+    public static class DictionaryExtension
+    {
+        /// <summary>
+        /// Dispose all elements of a <see cref="Dictionary{TKey,TValue}"/> and then clear the dictionary.
+        /// </summary>
+        /// <param name="dictionary">The <see cref="Dictionary{TKey,TValue}"/> to operate on.</param>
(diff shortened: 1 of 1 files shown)


jkeon (approved):

Collections/DictionaryExtension.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public static class DictionaryExtension
+    {
+        /// <summary>
+        /// Dispose all elements of a <see cref="Dictionary{TKey,TValue}"/> and then clear the dictionary.
+        /// </summary>
+        /// <param name="dictionary">The <see cref="Dictionary{TKey,TValue}"/> to operate on.</param>
+        /// <typeparam name="TKey">The key type</typeparam>
+        /// <typeparam name="TValue">The value type</typeparam>
+        public static void DisposeAllValuesAndClear<TKey, TValue>(this Dictionary<TKey, TValue> dictionary) where TValue : IDisposable
```
tjvezina, line 14: This is neat, I don't think I've seen an extension function with a `where` constraint - does that mean it only shows up in intellisense for the appropriate objects, or still all Dictionaries?

mbaker3: That's right. Or at least that's how Rider behaves!

PR #95, opened 2022-08-02
CollectionExtension - Fix error on DisposeAndClear

mbaker3:
Fix bug where `CollectionExtension.DisposeAndClear` would throw an exception on collections that cannot be cleared (ex: `Array`)

Non-clearable collections (Ex: Array) throw an error.

 - Add a guard against clearing collections where `ICollection<T>.IsReadOnly` is true. 
 - Rename method to reflect change in behaviour.

 - None

 - None

 - [x] Yes - Any uses of `CollectionExtension.DisposeAllAndClear` must be renamed to `DisposeAllAndTryClear`
 - [ ] No

### DIFF
--- a/Collections/CollectionExtension.cs
+++ b/Collections/CollectionExtension.cs
@@ -13,14 +13,17 @@ namespace Anvil.CSharp.Collections
         /// </summary>
         /// <param name="collection">The <see cref="ICollection{T}"/> to operate on.</param>
         /// <typeparam name="T">The element type</typeparam>
-        public static void DisposeAllAndClear<T>(this ICollection<T> collection) where T : IDisposable
+        public static void DisposeAllAndTryClear<T>(this ICollection<T> collection) where T : IDisposable
         {
             foreach (T item in collection)
             {
                 item.Dispose();
             }
 
-            collection.Clear();
+            if (!collection.IsReadOnly)
+            {
+                collection.Clear();
+            }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


Collections/CollectionExtension.cs
```diff
@@ -13,14 +13,17 @@ public static class CollectionExtension
         /// </summary>
         /// <param name="collection">The <see cref="ICollection{T}"/> to operate on.</param>
         /// <typeparam name="T">The element type</typeparam>
-        public static void DisposeAllAndClear<T>(this ICollection<T> collection) where T : IDisposable
+        public static void DisposeAllAndTryClear<T>(this ICollection<T> collection) where T : IDisposable
```
jkeon, line 16: I'm not sure about the naming here. I think we should keep it as `DisposeAllAndClear` 

To me the `Try` aspect is a runtime attempt that could change based on state. But in this case, it's really compile time, we just don't know because of generics. A ReadOnly list or an Array will ALWAYS fail. A regular list or hashset will ALWAYS succeed.

Seeing `DisposeAllAndTryClear` in code i haven't written gives me a bit of uncertainty. Do I need to check to make sure? 

Maybe just adding a `Remarks` section to the docs would be better?

jkeon (changes requested):
Looks good, just one change/thought. Open to being argued against though if you feel strongly the other way.

Collections/CollectionExtension.cs
mbaker3: For all of the examples I know of, you're right, `IsReadOnly` is locked in at compile time but the interface doesn't guarantee that.

As a developer consuming this method I hope you know whether the collection you're using can actually clear. That's why I'm not so worried about the try aspect.

Adding <remarks> is a good idea. I'll do that.

jkeon (approved):

PR #96, opened 2022-08-06
Changed Logging library .NET Framework version from 4.8 to 4.7.1

tjvezina:
Corresponding Unity PR: https://github.com/decline-cookies/anvil-unity-core/pull/69

A harmless change the Logging library's .NET version (it does not rely on any Framework 4.8-specific features) to satisfy Omnisharp in VS Code.

It seems Visual Studio and other IDE's are somehow smart enough to still use the Logging DLL despite the minor version mismatch, while Omnisharp just flatly gives up and stops loading the project until you fix it.

(Also corrected a cref from `Log.Logger` -> `Logger`)

Anvil.CSharp.Logging.dll targets .NET Framework 4.8, causing VS Code's C# extension to throw errors about using libraries with a newer .NET version than the project itself (Framework 4.7.1).

Anvil.CSharp.Logging.dll targets .NET Framework 4.7.1, allowing VS Code to open the project without complaint.

Unity projects using Anvil would fail to load in VS Code until you picked some arbitrary fix, like manually editing the .csproj to use 4.8, or giving up and using Visual Studio.

 - None

 - [ ] Yes
 - [X] No

### DIFF
--- a/Core/AbstractAnvilBase.cs
+++ b/Core/AbstractAnvilBase.cs
@@ -21,7 +21,7 @@ namespace Anvil.CSharp.Core
 
         private Logger? m_Logger;
         /// <summary>
-        /// Returns a <see cref="Log.Logger"/> for this instance to emit log messages with.
+        /// Returns a <see cref="Logger"/> for this instance to emit log messages with.
         /// Lazy instantiated.
         /// </summary>
         protected Logger Logger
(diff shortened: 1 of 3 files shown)


mbaker3 (approved):
Good catch. Looks good. Thanks!

PR #97, opened 2022-08-08
Add IDProvider

mbaker3:
Add `IDProvider` which supplies unique IDs with a warning emitted when the supply is running low.
 - Update `UpdateHandle` to use `IDProvider`
 - Minor documentation formatting fixes.

Each system that requires fast, simple, unique ID generation is left to implement the logic themselves. Leading to ad-hoc handling or no handling of ID supply exhaustion. 

 `IDProvider` consolidates the logic to one implementation that is less error prone and can include more features in the future.

By using a single `IDProvider` applications can now easily identify when ID supply is running out at a global level and react accordingly. This is an extreme scenario but one that some applications must watch out for.
**Note:** IDs aren't shared between provider instances.

None - This was ported over from an existing application and upgraded at the same time.

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Data/IDProvider.cs
+++ b/Data/IDProvider.cs
@@ -5,7 +5,7 @@ using Anvil.CSharp.Logging;
 namespace Anvil.CSharp.Data
 {
     /// <summary>
-    /// An instance that provides a unique IDs each time requested.
+    /// An instance that provides a unique ID each time it is requested.
     /// Includes a mechanism to detect when ID supply is near exhaustion.
     /// </summary>
     public class IDProvider
--- a/Data/IDProvider.cs
+++ b/Data/IDProvider.cs
@@ -45,7 +45,7 @@ namespace Anvil.CSharp.Data
         /// (default: <see cref="uint.MaxValue"/>) The threshold which, when passed, triggers
         /// <see cref="OnIDLimitImminent"/>.
         /// </param>
-        public IDProvider(uint supplyWarningThreshold = uint.MaxValue-1)
+        public IDProvider(uint supplyWarningThreshold = uint.MaxValue - 1_000_000)
         {
             SupplyWarningThreshold = supplyWarningThreshold;
         }
--- a/Data/IDProvider.cs
+++ b/Data/IDProvider.cs
@@ -11,11 +11,49 @@ namespace Anvil.CSharp.Data
     public class IDProvider
     {
         /// <summary>
+        /// Arguments that describe an ID limit warning.
+        /// </summary>
+        public class IDLimitWarningEventArgs : EventArgs
+        {
+            /// <summary>
+            /// Indicates whether the limit warning has been handled. If true, no further corrective action is required
+            /// by the application.
+            /// </summary>
+            public bool IsHandled { get; private set; }
+
+            /// <summary>
+            /// Creates a new instance.
+            /// </summary>
+            internal IDLimitWarningEventArgs()
+            {
+                IsHandled = false;
+            }
+
+            /// <summary>
+            /// Call if the limit has been handled and no corrective action is required by the application.
+            /// The event will continue to be raised to handlers for informational purposes (Logging, analytics, etc..)
+            /// </summary>
+            /// <remarks>
+            /// Calling <see cref="Handle"/> multiple times on a single event will raise a warning. While handling the
+            /// limit multiple times probably isn't harmful, it's not recommended.
+            /// </remarks>
+            public void Handle()
+            {
+                if (IsHandled)
+                {
+                    Log.GetLogger(this).Warning($"ID supply limit has already been handled by another listener and shouldn't get handled again.");
+                }
+
+                IsHandled = true;
+            }
+        }
+
+        /// <summary>
         /// Triggers the first time that any instance of <see cref="IDProvider"/> passes its
         /// <see cref="SupplyWarningThreshold"/> for the first time.
         /// This gives the application the opportunity to react before IDs are exhausted.
         /// </summary>
-        public static event Action<IDProvider> OnIDLimitGlobalWarning;
+        public static event EventHandler<IDLimitWarningEventArgs> OnIDLimitGlobalWarning;
 
         /// <summary>
         /// A reserved ID to represent "No ID".
@@ -27,7 +65,7 @@ namespace Anvil.CSharp.Data
         /// Triggers the first time that <see cref="SupplyWarningThreshold"/> is passed.
         /// This gives the consumer the opportunity to react before IDs are exhausted.
         /// </summary>
-        public event Action OnIDLimitWarning;
+        public event EventHandler<IDLimitWarningEventArgs> OnIDLimitWarning;
 
         /// <summary>
         /// The threshold which, when passed, triggers <seealso cref="OnIDLimitWarning"/>.
@@ -75,13 +113,17 @@ namespace Anvil.CSharp.Data
 
         private void CheckIfIDThresholdCrossed()
         {
-            if (m_LastNewID > SupplyWarningThreshold && m_HasIDWarningTriggered)
+            if (m_LastNewID <= SupplyWarningThreshold || m_HasIDWarningTriggered)
             {
-                Log.GetLogger(this).Warning($"{nameof(IDProvider)} has passed its supply warning threshold. Threshold: {SupplyWarningThreshold}");
-                m_HasIDWarningTriggered = true;
-                OnIDLimitWarning?.Invoke();
-                OnIDLimitGlobalWarning?.Invoke(this);
+                return;
             }
+
+            Log.GetLogger(this).Warning($"{nameof(IDProvider)} has passed its supply warning threshold. Threshold: {SupplyWarningThreshold:N0}");
+            m_HasIDWarningTriggered = true;
+
+            IDLimitWarningEventArgs args = new IDLimitWarningEventArgs();
+            OnIDLimitWarning?.Invoke(this, args);
+            OnIDLimitGlobalWarning?.Invoke(this, args);
         }
     }
 }
--- a/Data/IDProvider.cs
+++ b/Data/IDProvider.cs
@@ -1,0 +1,84 @@
+﻿using System;
+using System.Diagnostics;
+using Anvil.CSharp.Logging;
+
+namespace Anvil.CSharp.Data
+{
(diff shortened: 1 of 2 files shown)


Data/IDProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+
+
+        /// <summary>
+        /// Creates a new ID provider optionally allowing a supply warning threshold to be set.
+        /// </summary>
+        /// <param name="supplyWarningThreshold">
+        /// (default: <see cref="uint.MaxValue"/>) The threshold which, when passed, triggers
+        /// <see cref="OnIDLimitImminent"/>.
+        /// </param>
+        public IDProvider(uint supplyWarningThreshold = uint.MaxValue-1)
```
tjvezina, line 45:
```suggestion
        public IDProvider(uint supplyWarningThreshold = uint.MaxValue-1_000_000)
```
On the one hand, any threshold is exceedingly unlikely to be hit - but on the other hand, why not leave a huge amount of wiggle room by default? For any arbitrary system, one million ID's remaining seems like plenty of time to refresh the system, but you still have to burn through 99.977% of ID's to get there!

tjvezina (approved):
Nice solution, cheap and lightweight ID's, vanishingly unlikely to run out!

Just one suggestion, but not an important one if you want to ignore it.

Data/IDProvider.cs
```diff
@@ -0,0 +1,84 @@
+﻿using System;
+using System.Diagnostics;
+using Anvil.CSharp.Logging;
+
+namespace Anvil.CSharp.Data
+{
+    /// <summary>
+    /// An instance that provides a unique IDs each time requested.
```
jkeon, line 8:
```suggestion
    /// An instance that provides a unique ID each time it is requested.
```

Data/IDProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// <see cref="SupplyWarningThreshold"/> for the first time.
+        /// This gives the application the opportunity to react before IDs are exhausted.
+        /// </summary>
+        public static event Action<IDProvider> OnIDLimitGlobalWarning;
+
+        /// <summary>
+        /// A reserved ID to represent "No ID".
+        /// <see cref="GetNextID"/> will never return this value even when IDs roll over and are no longer unique.
+        /// </summary>
+        public const int UNSET_ID = 0;
```
jkeon, line 24: Should this be uint?

Data/IDProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// Includes a mechanism to detect when ID supply is near exhaustion.
+    /// </summary>
+    public class IDProvider
+    {
+        /// <summary>
+        /// Triggers the first time that any instance of <see cref="IDProvider"/> passes its
+        /// <see cref="SupplyWarningThreshold"/> for the first time.
+        /// This gives the application the opportunity to react before IDs are exhausted.
+        /// </summary>
+        public static event Action<IDProvider> OnIDLimitGlobalWarning;
```
jkeon, line 18: Hmmmn... this seems like the kind of thing that I might do that you would then call me out on :D

I'm not sure this should be static. If you create an `IDProvider` somewhere you should listen to that instance and handle it there too rather than having a global handler for any or all `IDProvider`s. Especially if someone else (different assembly, different third party mod etc) is handling this event before you get to and doing something they shouldn't.

jkeon: Oh ok, I just saw the instance event below... I'm still not sure we want to expose this though. But maybe i'm missing a use case?

Data/IDProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// </summary>
+        public const int UNSET_ID = 0;
+
+        /// <summary>
+        /// Triggers the first time that <see cref="SupplyWarningThreshold"/> is passed.
+        /// This gives the consumer the opportunity to react before IDs are exhausted.
+        /// </summary>
+        public event Action OnIDLimitWarning;
+
+        public readonly uint SupplyWarningThreshold;
```
jkeon, line 32: docs?

Data/IDProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            uint id = ++m_LastNewID;
+            CheckIfIDThresholdCrossed();
+
+            if (m_LastNewID == 0)
+            {
+                Log.GetLogger(this).Error($"{nameof(IDProvider)} has passed its maximum ID value. IDs provided are no longer guaranteed to be unique.");
+                // Push the ID past the UNSET_ID value so at least THAT remains unique.
+                id = ++m_LastNewID;
+            }
```
jkeon, line 67: Does this work? I thought you had to use the `checked` or `unchecked` when dealing with overflow?

DelayedExecution/UpdateHandle.cs
```diff
@@ -39,9 +38,11 @@ public class UpdateHandle : AbstractAnvilBase
...
         }
 
-
-        private uint m_CallAfterHandleCurrentID = CALL_AFTER_HANDLE_INITIAL_ID;
-
+        // Arbitrary choice to warn when there are only 100 IDs left to provide.
+        // Seems likely that an application wouldn't need more than 100 more delayed calls before it can
+        // create a new UpdateHandle to work with.
+        // If this is an issue there's no risk to having the threshold trigger sooner.
+        private readonly IDProvider m_IDProvider = new IDProvider(UInt32.MaxValue-100);
```
jkeon, line 45:
```suggestion
        private readonly IDProvider m_IDProvider = new IDProvider(uint.MaxValue-100);
```

jkeon (changes requested):
Looks good, just a few comments

Data/IDProvider.cs
jkeon: That seems reasonable to me

Data/IDProvider.cs
mbaker3: This is here as a high level application catch all. Rather than having to dig up every instance used you can listen to one static event. The only solution we know of at the moment is a soft reset and `uint` is such a large ID pool that it's an infrequent case. I think most applications will just want to know, at a high level, is some ID pool running out? Reload the whole application.

If you are writing an application that wants to only handle high risk `IDProviders` or specific subsets then you can still do that with the instance level event.

As a developer do you really want to go through and figure out everywhere an `IDProvider` is used?
Or have to design every Anvil system that uses an `IDProvider` indicate this edge case?

A good example is `UpdateHandle`. Everytime we use `UpdateHandle` do we now have to listen for the event on each instance?

If someone used this event anywhere in Anvil I'd reject it. The idea is that one place in an application listens to this.

I'm open to other suggestions but to me this is the right balance of safety and effort vs likelihood.

Data/IDProvider.cs
mbaker3: Using those keywords overrides the compiler setting. By default Unity doesn't have arithmetic overflow checking enabled. Even with Editor Debug mode enabled. I'm not sure how you turn on checking application wide in Unity, it is a compiler argument that you'd normally enable through project preferences in the IDE.

(Isn't enabled in .Net Fiddle either https://dotnetfiddle.net/58H0eP)

I'm thinking we leave it as is. If somebody turns on arithmetic checking then they'll get an exception. Since the overflow of the ID is an error case, that we've just provided graceful handling for, it makes sense to let it throw an exception if the developer wanted the safety of overflow checking.

DelayedExecution/UpdateHandle.cs
mbaker3: oops, weird!

Data/IDProvider.cs
mbaker3: yep, fixed

Data/IDProvider.cs
mbaker3: Fixed

Data/IDProvider.cs
mbaker3: Sure, 1MM is actually what I chose for the threshold in the game. I was thinking that these would generally be created in a much more rigid context that was able to abort more quickly but you make a good point. I've removed the parameter from `UpdateHandle`'s usage of `IDProvider`

Fixed

mbaker3:
PR Feedback addressed. Ready for re-review.
Thanks!

Data/IDProvider.cs
jkeon: Yep, ok this makes sense.

Data/IDProvider.cs
jkeon: Right, ok that makes sense. 

I wonder if we should do the following then:

1. Get rid of the instance event, and have it be an optional `Action` in the constructor. 

You either say "yes, we can gracefully handle this, or we know what we have to do to handle this and we will. No one else should interfere" OR you say "we can't handle this, it's up the higher level to do so via the static event."

2. If the instance Action is not null, then fire that and DON'T fire the static event. Upon creation we've stated we can and will handle it. If the instance Action is null, then fire the static event. The upper level can handle it if it chooses to. 

I think this makes the API safer and better deals with the intent. In `UpdateHandle`, we can't gracefully handle it. It's a signal for the app to soft restart. The static event should fire and app devs can handle it. 

But if we end up having something in the future that we can handle, we'll know that at creation and have passed in the instance Action to do something to handle it. I wouldn't want the app to get that static event and decide to soft restart since we've already coded the handling.

jkeon (changes requested):
Looks good, just going a bit further on the events. Let me know what you think.

Data/IDProvider.cs
mbaker3: That's a good point. I don't love the idea of submitting a callback into `IDProvider` but we could accomplish the same with the event and args that allow a handler to mark the event as "handled".

That would also allow others know about the warning occurring without necessarily needing to handle it. Since exhausted ID supply handling will be expensive in most cases I can see applications wanting to know when the operation has occurred and logging an event to analytics.

mbaker3:
Adjusted limit warnings to allow a handler to mark the warning as handled. This allows corrective action to be taken before hitting the application scope but still allows the application to be aware that the warning was reached (even if handled gracefully).

Also snuck in some minor changes
 - Improved the formatting of the log messages with uint values
 - Reduced nesting in the threshold check method

jkeon (approved):
Looks good.

Issue #105, opened 2022-08-17
Write Unit Tests for Array2DExtension


mbaker3:
Depends On: #106 

Write unit tests for the `Array2DExtension` class

**Here's a brief start**
These are the half assed tests that were run when first writing the extension.

``` csharp
bool[,] test = new bool[5,5];
test.Populate((x,y)=> true);

Debug.LogError(test.All((val, x, y) => val == true)); // true
Debug.LogError(test.All((val, x, y) => val == false)); // false
Debug.LogError(test.Any((val, x, y) => val == false)); // false

test[2, 2] = false;
Debug.LogError(test.All((val, x, y) => val == true)); // false
Debug.LogError(test.Any((val, x, y) => val == true)); // true
Debug.LogError(test.Any((val, x, y) => val == false)); // true
```

Issue #104, opened 2022-08-17
Optimize `Array2DExtension`


mbaker3:
Related: 

Performance optimize the operations in `Array2DExtensions`

PR #107, opened 2022-08-17
Anvil2DExtension - Add

mbaker3:
Add a collection of extensions for working with 2 dimensional arrays.
This is a quick implementation that is being migrated from an existing project. 
Follow up tasks to improve this class are: #104 #105 

Working with 2D arrays is cumbersome.

Working with 2D arrays is more pleasant.

None

 - None (Except the PR in the project this was pulled from)

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Data/Collections/Util/Array2DExtension.cs
+++ b/Data/Collections/Util/Array2DExtension.cs
@@ -1,0 +1,161 @@
+﻿using System;
+
+namespace Anvil.CSharp.Data
+{
+    //TODO: #104 - Optimize
+    /// <summary>
+    /// A set of helper extension methods for 2D arrays
+    /// </summary>
+    public static class Array2DExtension
+    {
+        /// <summary>
+        /// Gets the value at the specified index or returns the default value if out of range.
+        /// </summary>
+        /// <param name="array">The array to query.</param>
+        /// <param name="x">The first order index</param>
+        /// <param name="y">The second order index</param>
+        /// <typeparam name="T">The type of the array.</typeparam>
+        /// <returns>The value at the specified index or default.</returns>
+        public static T GetElementOrDefaultAt<T>(this T[,] array, int x, int y)
+        {
+            int limitX = array.GetLength(0);
+            int limitY = array.GetLength(1);
+
+            if (x >= limitX || y >= limitY || x < 0 || y < 0)
+            {
+                return default;
+            }
+
+            return array[x, y];
+        }
+
+        /// <summary>
+        /// Gets the index of the first element in the array that satisfies the specified <see cref="Predicate{T}"/>.
+        /// </summary>
+        /// <param name="array">The array to query.</param>
+        /// <param name="predicate">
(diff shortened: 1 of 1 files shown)


Data/Collections/Util/Array2DExtension.cs
```diff
@@ -0,0 +1,161 @@
...
+        /// Gets the index of the first element in the array that satisfies the specified <see cref="Predicate{T}"/>.
+        /// </summary>
+        /// <param name="array">The array to query.</param>
+        /// <param name="predicate">
+        /// The <see cref="Predicate{T}"/> to call for each element. If the predicate returns true the method returns
+        /// the current index.
+        /// </param>
+        /// <typeparam name="T">The type of the array.</typeparam>
+        /// <returns>The first index that satisfied the predicate.</returns>
+        public static (int,int) IndexOf<T>(this T[,] array, Predicate<T> predicate)
```
tjvezina, line 42: I would suggest another name (maybe `FindIndexOf<T>`?) to indicate a conditional search - `IndexOf<T>` makes me think I'm just passing an element/value and asking for its index in the array.

Data/Collections/Util/Array2DExtension.cs
```diff
@@ -0,0 +1,161 @@
...
+    {
+        /// <summary>
+        /// Gets the value at the specified index or returns the default value if out of range.
+        /// </summary>
+        /// <param name="array">The array to query.</param>
+        /// <param name="x">The first order index</param>
+        /// <param name="y">The second order index</param>
+        /// <typeparam name="T">The type of the array.</typeparam>
+        /// <returns>The value at the specified index or default.</returns>
+        public static T GetElementOrDefaultAt<T>(this T[,] array, int x, int y)
```
tjvezina, line 19: Would an optional explicit default value be useful? Mostly for primitives I think, so you could say `array.GetElementOrDefaultAt(x, y, -1);`

tjvezina (commented):
I don't know if this PR was a secret, or you just forgot to request reviews, but I found it!

Data/Collections/Util/Array2DExtension.cs
mbaker3: Good call. I'll follow the lead of `Linq` and `Array` and name the method `FindIndex`

mbaker3:
> I don't know if this PR was a secret, or you just forgot to request reviews, but I found it!

lol, not a secret. I think you can watch the repo if you really want notifications for every PR. This didn't touch logging so I didn't ping you!

I can ping you for everything if youwant!

tjvezina (approved):
I can think of more potentially helpful extensions, but I'm sure they'll be added as use cases come up.

tjvezina:
> I can ping you for everything if you want!

Huh, I guess I was thinking of the old pr-notify group (or whatever it was called), because I normally do get emails about all new PR's, but this one slipped by. I'll check my settings, don't worry about it!

PR #109, opened 2022-09-02
AbstractIDProviders with Uint and Byte implementations

jkeon:
Refactored the `IDProvider` into an `AbstractIDProvider` and provided specific implementations for `uint` and `byte`.

Currently the `IDProvider` assumes a backing type of `uint` and there are no other options. This can represent a range of ID's that is rather large and may not be needed. Each ID is 4 bytes which may be excessive memory wise.

The `IDProvider` has been abstracted so that any amount of types can be created. Unfortunately we cannot just have a generic implementation until C#11/.Net6 because there's no way to do mathematical operations on generics. 

Instead I've implemented a `UintIDProvider` which is the same as `IDProvider` and a `ByteIDProvider` which uses a `byte` as the backing type. Much smaller range but each ID is only 1 byte. 

Creating new types is pretty easy and painless to do. Just need to implement a function for the increment and greater than check. I opted to not implement for every data type `ushort` and `ulong` until it is necessary.

Note that the global warning event will be raised regardless of which `AbstractIDProvider` gets triggered. This way you don't have to subscribe to a static event for every "type" in your project. 

- None

 - None

 - [x] Yes - Only if you are already using an `IDProvider`, you'll just have to use the new `UintProvider` class instead. The API is the same.
 - [ ] No


DelayedExecution/UpdateHandle.cs
```diff
@@ -42,7 +42,7 @@ public class UpdateHandle : AbstractAnvilBase
         // Seems likely that an application wouldn't need more than 100 more delayed calls before it can
         // create a new UpdateHandle to work with.
         // If this is an issue there's no risk to having the threshold trigger sooner.
-        private readonly IDProvider m_IDProvider = new IDProvider();
+        private readonly UintIDProvider m_IDProvider = new UintIDProvider();
```
tjvezina, line 45: I think this comment is outdated, from when this IDProvider instance had an explicit 100-ID's-left warning

tjvezina (approved):
Interesting that it's worth saving a few bytes per entity! On the one hand, I'm sure it adds up with volume, but on the other hand you can only generate ~200 byte ID's per system, so it doesn't seem like a huge savings? Of course, I have no context, and nothing against these changes, just thinking out loud!

Is there any concern about a system using `ByteIDProvider` and reaching it's small, arguably less dire limit, tripping the global warning and causing a heavy-handed reaction like reloading the whole game? I imagine you'd find out about situations like that during development anyway.

jkeon:
> Interesting that it's worth saving a few bytes per entity! On the one hand, I'm sure it adds up with volume, but on the other hand you can only generate ~200 byte ID's per system, so it doesn't seem like a huge savings? Of course, I have no context, and nothing against these changes, just thinking out loud!

Very fair point. In this case it's less about the total memory being used and just wanting to get it small enough so that a mapping of enums to pointers fits in a 64 byte cache line. There's more context here: https://github.com/decline-cookies/anvil-unity-dots/pull/52#issuecomment-1234602907 in that comment and the overall PR but that PR is a bit more loose since it's merging into a temp branch in order to split up a monolith feature. 

> Is there any concern about a system using ByteIDProvider and reaching it's small, arguably less dire limit, tripping the global warning and causing a heavy-handed reaction like reloading the whole game? I imagine you'd find out about situations like that during development anyway.

There definitely is but I think that it's on you as the developer to choose a provider that gives you reasonable chances that you're not going to hit it. If in doubt, use a larger backing type but if you know there will only be a handful, a `byte` gives you loads of room.

DelayedExecution/UpdateHandle.cs
jkeon: Good point. I've removed the comment completely. The defaults in the providers are enough to go on.

Data/AbstractIDProvider.cs
```diff
@@ -1,14 +1,95 @@
...
+        {
+            if (!HasIDExceededSupplyWarningThreshold(m_LastNewID)|| m_HasIDWarningTriggered)
+            {
+                return;
+            }
+
+            Log.GetLogger(this).Warning($"{nameof(AbstractIDProvider<T>)} has passed its supply warning threshold. Threshold: {SupplyWarningThreshold:N0}");
+            m_HasIDWarningTriggered = true;
+
+            IDLimitWarningEventArgs args = new AbstractIDProvider.IDLimitWarningEventArgs();
```
mbaker3, line 82: Redundant qualifier

```suggestion
            IDLimitWarningEventArgs args = new IDLimitWarningEventArgs();
```

mbaker3 (changes requested):
Nice change. 
I don't love the having the generic and non-generic variants but I understand the necessity.

Thinking out loud, would it be better to have a separate `GlobalIDProviderDispatcher` or `IDProviderGlobalDispatcher` object that provides the one place to listen to the global ID limit event?
 - Right now it's a bit weird to listen to a subtype but get events for all types (Ex: `ByteIDProvider.OnIDLimitGlobalWarning`). I could see a developer accidentally listening to the global event from multiple types.
 - IDProviderGlobalDispatcher would eliminate the need for a non-generic abstract class.
 
 Alternatively maybe we made the typed providers inner classes to a static `ID` type?
 ```csharp
ID.Provider myProvider = new ID.Provider();
ID.ByteProvider myByteProvider = new ID.ByteProvider();

ID.OnIDLimitGlobalWarning += myHandler;
```

Just spit balling. Don't need to try a significant change. Just ideas that came up while trying to get over the discomfort of the current solution.

File: Data/AbstractIDProvider.cs [hunk 3]
```diff
@@ -1,14 +1,95 @@
...
+        /// If <typeparamref name="T"/> has rolled over an error will be logged and IDs are no longer guaranteed to be unique.
+        /// </remarks>
+        public T GetNextID()
+        {
+            m_LastNewID = IncrementID(m_LastNewID);
+            CheckIfIDThresholdCrossed();
+
+            if (m_LastNewID.Equals(default))
+            {
+                Log.GetLogger(this).Error($"{nameof(AbstractIDProvider<T>)} has passed its maximum ID value. IDs provided are no longer guaranteed to be unique.");
```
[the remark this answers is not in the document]
jkeon: Fixed.

Data/AbstractIDProvider.cs
```diff
@@ -47,83 +128,17 @@ public void Handle()
...
+        
         /// <summary>
-        /// Triggers the first time that any instance of <see cref="IDProvider"/> passes its
-        /// <see cref="SupplyWarningThreshold"/> for the first time.
+        /// Triggers the first time that any instance of <see cref="AbstractIDProvider"/> passes its
+        /// SupplyWarningThreshold for the first time.
         /// This gives the application the opportunity to react before IDs are exhausted.
         /// </summary>
-        public static event EventHandler<IDLimitWarningEventArgs> OnIDLimitGlobalWarning;
+        internal static event EventHandler<IDLimitWarningEventArgs> Internal_OnIDLimitGlobalWarning;
```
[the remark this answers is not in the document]
jkeon: No longer relevant with ID refactor

Data/ByteIDProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// <summary>
+    /// Specific implementation of <see cref="AbstractIDProvider{T}"/> for <see cref="byte"/>
+    /// </summary>
+    /// <inheritdoc cref="AbstractIDProvider{T}"/>
+    public class ByteIDProvider : AbstractIDProvider<byte>
+    {
+        /// <summary>
+        /// Triggers the first time that any instance of <see cref="ByteIDProvider"/> passes its
+        /// <see cref="AbstractIDProvider{T}.SupplyWarningThreshold"/> for the first time.
+        /// This gives the application the opportunity to react before IDs are exhausted.
```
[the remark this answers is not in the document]
jkeon: No longer relevant with ID refactor

Data/UintIDProvider.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+
+namespace Anvil.CSharp.Data
+{
+    
+    /// <summary>
+    /// Specific implementation of <see cref="AbstractIDProvider{T}"/> for <see cref="uint"/>
+    /// </summary>
+    /// <inheritdoc cref="AbstractIDProvider{T}"/>
+    public class UintIDProvider : AbstractIDProvider<uint>
```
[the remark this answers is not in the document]
jkeon: Done

Data/UintIDProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        /// <summary>
+        /// Triggers the first time that any instance of <see cref="UintIDProvider"/> passes its
+        /// <see cref="AbstractIDProvider{T}.SupplyWarningThreshold"/> for the first time.
+        /// This gives the application the opportunity to react before IDs are exhausted.
```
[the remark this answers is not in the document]
jkeon: No longer relevant with the ID refactor

jkeon:
I reworked this a bit based on the feedback and I agree, I think this is cleaner. 

You create your `IDProvider` or `ByteIDProvider` as you normally would. It will register itself for the static event and dispatch when the instance event is called.

One thing I noticed when making this is that there was no disposal. I added that to make sure instances get cleaned up properly and the static event is also cleaned up but the onus is on the developer to call the static `Dispose` on a soft-reload in app. 

I've made a Unity PR to handle domain reloading automatically.

Data/ID/ID.cs
```diff
@@ -0,0 +1,93 @@
...
+            ID_PROVIDERS.Add(provider);
+            provider.OnIDLimitWarning += Provider_OnIDLimitWarning;
+        }
+
+        internal static void UnregisterIDProvider(AbstractIDProvider provider)
+        {
+            Debug.Assert(ID_PROVIDERS.Contains(provider), $"{provider} is being unregistered from {nameof(ID)} but it wasn't registered!");
+            ID_PROVIDERS.Remove(provider);
+            provider.OnIDLimitWarning -= Provider_OnIDLimitWarning;
+        }
```
mbaker3, line 86: This isn't going to behave the way we want it to. We want the instance limit listeners to have an opportunity to handle the limit reached before the global event kicks off.

That behaviour allows local providers to potentially recover from supply exhaustion before the application is required to handle it at the top level.

Suggest we just create an `internal` notify global method and then we can just get rid of the need to maintain a list of provider instances.

Data/ID/ID.cs
```diff
@@ -0,0 +1,93 @@
...
+        private static readonly HashSet<AbstractIDProvider> ID_PROVIDERS = new HashSet<AbstractIDProvider>();
+
+        /// <summary>
+        /// Removes all reference to any <see cref="AbstractIDProvider"/>s that were created and resets
+        /// the <see cref="OnIDLimitGlobalWarning"/> event.
+        /// </summary>
+        /// <remarks>
+        /// Useful for when the application is soft-reloaded from inside.
+        /// </remarks>
+        public static void Dispose()
```
mbaker3, line 64: Can we call this `Reset` since it's not really disposing?

mbaker3: This method may not be required based on other comments in this and the Unity PR

mbaker3 (changes requested):
On issue with the global event and then some cleanup recommendations.

Data/ID/ID.cs
jkeon: Hmmn... good point. Fixed

Data/ID/ID.cs
jkeon: Yeah I don't think its necessary anymore. Removed.

mbaker3 (approved):

Issue #108, opened 2022-08-31
FileLogHandler - Implement Rotating Log Handling and Culling


mbaker3:
When `FileLogHandler` is being used with the `append` parameter set to true the individual log file will grow indefinitely.

This does not necessarily need to be implemented in FileLogHandler. Whoever tackles this task can decide the best course.

Implement a mechanism to:
 - Move to a new log file when it reaches a specific size
 - Cull old log files

This should work similarly to [logrotate](https://www.tecmint.com/install-logrotate-to-manage-log-rotation-in-linux/) on linux.
 - The initial file name is as defined by the developer. `myLog.log`
 - When the initial log file fills up it should be renamed to `myLog.1.log` and a new `myLog.log` created.
 - When the log file fills up again all logs should be rotated by 1 and the current log file becoming the new `myLog.1.log.

This results in a log collection with a maximum rotation count of 3 looking like:
 - `myLog.log` - The log file currently being written to
 - `myLog.1.log` - The next newest log file where the last entry is the log emitted immediately before the first entry in `myLog.log`. When `myLog.log` fills up this will be renamed to `myLog.2.log`
 - `myLog.2.log`
 - `myLog.3.log` - The oldest log file. When `myLog.log` fills this file will be culled.

The consuming developer should have parameters exposed (with sane defaults) to configure:
 - Threshold size to roll over to a new log
 - Maximum number of times a log may be rotated.

tjvezina:
I don't see the harm in making these options available to non-appended logs too - maybe you've got a lot of logs going on and you'd like them rotated, even if they're wiped out and restarted each run.

That does complicate the implementation a bit though - would it be worth setting that up for the increased flexibility, or should it only apply to appended logs? In the latter case, I'm thinking a `RotatedFileLogHandler` derived from `FileLogHandler`.

tjvezina:
Also, in a completely biased move based on a single recent experience (Plex logs), I'm thinking the default limits could be 10MB / 5 files. Thoughts?

jkeon:
I would maybe take it further and say it's good policy to keep a history (however short) of the logs regardless. 

Whether it's append or not, in a situation where you need to look at the log, it's probably helpful to be able to look at the previous session's log as well for comparison. That might make the implementation simpler.

With the default limits, I don't have a strong opinion, I'm fine with the 10MB / 5 files default. I would imagine the API will make it trivial for each application to set those limits as it best suits them.

tjvezina:
[...]

PR #111, opened 2022-09-08
Logging - Log File Rotation

tjvezina:
Implemented log file rotation (see linked issue for details) in `FileLogHandler`.

Log files created by `FileLogHandler` may grow to indefinite size, especially with `append: true`.

With file rotation enabled (which it is by default), after a log file reaches the set size limit, it is "rotated" (i.e. renamed with an index suffix) and a new file started.

This prevents any one log file from becoming too large to reasonably work with.

 - None

 - [ ] Yes
 - [X] No
_Note: Any existing uses of `FileLogHandler` will automatically begin using file rotation._

---

_Example logs generated under various use cases:_
image

_Strict file size limit (rotate before the log that exceeds the limit)_
image

### Updated Test Cases
- Re-tested every append/replace/rotate combination (as per first screenshot above)
- `RotateFileSizeLimit` is null/negative/zero/`long.MaxValue`
- `RotateFileCountLimit` is null/negative/zero/`int.MaxValue`
- Path contains folders that don't exist
- Unusual paths:
  - `path` has no file extension (ex. "log" -> "log.1", "log.2"...)
  - `path` already has an index (ex. "log.1.txt" -> "log.1.1.txt", "log.1.2.txt"...)
    - _Index is treated as part of the file name_
  - `path` ends with an index (ex. "log.1" -> "log.1.1", "log.2.1")
    - _Index is treated as the file extension_
- User manually modifies log files:
  - Gaps in rotated file indices
  - Duplicate indices (ex "log.1.txt" vs "log.01.txt")
  - Indices greater than `int.MaxValue` (ex. "log.123456789123456789.txt")

### DIFF
--- a/Logging/LogHandler/FileLogHandler.cs
+++ b/Logging/LogHandler/FileLogHandler.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.IO;
-using System.Text;
 using Anvil.CSharp.Core;
 
 namespace Anvil.CSharp.Logging
@@ -10,12 +9,30 @@ namespace Anvil.CSharp.Logging
     /// </summary>
     public class FileLogHandler : AbstractAnvilBase, ILogHandler
     {
+        /// <summary>
+        /// The type of log files the handler should generate.
+        /// </summary>
+        public enum LogType
+        {
+            /// <summary>
+            /// New logs are appended to the existing log file.
+            /// </summary>
+            Append,
+            /// <summary>
+            /// A new log file is created each session, overwriting the previous one if it exists.
+            /// </summary>
+            Replace,
+        }
+
         public const string LOG_CONTEXT_CALLER_DERIVED_TYPE = "{0}";
         public const string LOG_CONTEXT_CALLER_FILE = "{1}";
         public const string LOG_CONTEXT_CALLER_METHOD = "{2}";
         public const string LOG_CONTEXT_CALLER_LINE = "{3}";
 
-        private readonly StreamWriter m_Writer;
+        private StreamWriter m_Writer;
+        private readonly string m_Path;
+        private readonly LogType m_LogType;
+        private readonly bool m_Rotate;
 
         /// <summary>
         /// Indicates whether to prefix logs with a timestamp.
@@ -57,13 +74,57 @@ namespace Anvil.CSharp.Logging
         public LogLevel MinimumLevel { get; set; } = LogLevel.Debug;
 
         /// <summary>
+        /// The maximum file size in bytes before the file is rotated. Default is 10MB.
+        /// </summary>
+        public long RotateFileSizeLimit { get; set; } = 10L * 1024 * 1024;
+
+        /// <summary>
+        /// The maximum number of rotated files to keep before deleting the oldest. Default is 5.
+        /// Ex. If the limit is 5, up to 5 rotated files in addition to the original log file, will be kept. After reaching this limit,
+        /// the next time the log file is rotated, the oldest file will be deleted. If the limit is 0, no rotated files are kept,
+        /// i.e. when the log file is rotated, it is simply erased and starts over.
+        /// </summary>
+        public int RotateFileCountLimit { get; set; } = 5;
+
+        /// <summary>
         /// Creates a `FileLogHandler` that can write to the given file path.
         /// </summary>
         /// <param name="path">The text file to output messages to.</param>
-        /// <param name="append">If true, the file will have new logs appended to it, otherwise the file is replaced.</param>
-        public FileLogHandler(string path, bool append = true)
+        /// <param name="logType">The <see cref="LogType"/> to use.</param>
+        /// <param name="rotate">
+        /// If true, the file will be rotated when the size limit is reached. When a file is rotated, it has an index appended to it,
+        /// i.e. "log.txt" is renamed "log.1.txt", and a new "log.txt" is opened. If "log.1.txt" already exists, it is first renamed
+        /// "log.2.txt", and so on, rotating through log files. Once the file count limit is reached, the oldest file is deleted.
+        /// </param>
+        public FileLogHandler(string path, LogType logType = LogType.Append, bool rotate = true)
         {
-            m_Writer = new StreamWriter(path, append)
+            m_Path = path;
+            m_LogType = logType;
+            m_Rotate = rotate;
+
+            string directory = Path.GetDirectoryName(m_Path);
+            if (!Directory.Exists(directory))
+            {
+                Directory.CreateDirectory(directory);
+            }
+
+            CreateWriter();
+
+            if (logType == LogType.Replace && rotate)
+            {
+                int index = 1;
+                string rotatedFilePath = GetRotatedFilePath(index);
+                while (File.Exists(rotatedFilePath))
+                {
+                    File.Delete(rotatedFilePath);
+                    rotatedFilePath = GetRotatedFilePath(++index);
+                }
+            }
+        }
+
+        private void CreateWriter()
+        {
+            m_Writer = new StreamWriter(m_Path, append: (m_LogType == LogType.Append))
             {
                 AutoFlush = true
             };
@@ -97,6 +158,56 @@ namespace Anvil.CSharp.Logging
                 );
 
             m_Writer.WriteLine($"{timestamp}{logLevel}{context}{message}");
+
+            if (m_Rotate && m_Writer.BaseStream.Length > RotateFileSizeLimit)
+            {
+                this.RotateFiles();
+            }
+        }
+
+        private void RotateFiles()
+        {
+            m_Writer.Close();
+            m_Writer = null;
+
+            // Determine how many rotated log files already exist
+            int index = 1;
+            string rotatedFilePath = GetRotatedFilePath(index);
+            while (File.Exists(rotatedFilePath))
+            {
+                rotatedFilePath = GetRotatedFilePath(++index);
+            }
+            --index;
+
+            // Delete any files in excess of the file count limit
+            while (index >= RotateFileCountLimit)
+            {
+                File.Delete(GetRotatedFilePath(index));
+                --index;
+            }
+
+            // Rotate remaining files
+            while (index >= 0)
+            {
+                File.Move(GetRotatedFilePath(index), GetRotatedFilePath(index + 1));
+                --index;
+            }
+
+            CreateWriter();
+        }
+
+        private string GetRotatedFilePath(int index)
+        {
+            if (index == 0)
+            {
+                return m_Path;
+            }
+
+            string directory = Path.GetDirectoryName(m_Path);
+            string fileName = Path.GetFileNameWithoutExtension(m_Path);
+            string extension = Path.GetExtension(m_Path);
+            
+            return Path.Combine(directory, $"{fileName}.{index}{extension}");
         }
     }
 }
--- a/Logging/LogHandler/FileLogHandler.cs
+++ b/Logging/LogHandler/FileLogHandler.cs
@@ -19,7 +19,7 @@ namespace Anvil.CSharp.Logging
             /// </summary>
             Append,
             /// <summary>
-            /// A new log file is created each session, overwriting the previous one if it exists.
+            /// A new log file is created, overwriting the previous one if it exists.
             /// </summary>
             Replace,
         }
--- a/Logging/LogHandler/FileLogHandler.cs
+++ b/Logging/LogHandler/FileLogHandler.cs
@@ -1,6 +1,9 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.IO;
-using System.Diagnostics;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
 using Anvil.CSharp.Core;
 
 namespace Anvil.CSharp.Logging
@@ -13,7 +16,7 @@ namespace Anvil.CSharp.Logging
         /// <summary>
         /// The type of log files the handler should generate.
         /// </summary>
-        public enum LogType
+        public enum WriteMode
         {
             /// <summary>
             /// New logs are appended to the existing log file.
@@ -36,8 +39,10 @@ namespace Anvil.CSharp.Logging
         private readonly string m_FileName;
         private readonly string m_Extension;
 
-        private readonly LogType m_LogType;
-        private readonly bool m_ShouldRotate;
+        private readonly WriteMode m_WriteMode;
+        
+        private readonly long? m_RotateFileSizeLimit;
+        private readonly int? m_RotateFileCountLimit;
 
         /// <summary>
         /// Indicates whether to prefix logs with a timestamp.
@@ -79,33 +84,42 @@ namespace Anvil.CSharp.Logging
         public LogLevel MinimumLevel { get; set; } = LogLevel.Debug;
 
         /// <summary>
-        /// The maximum file size in bytes before the file is rotated. Default is 10MB.
-        /// </summary>
-        public long RotateFileSizeLimit { get; set; } = 10L * 1024 * 1024;
-
-        /// <summary>
-        /// The maximum number of rotated files to keep before deleting the oldest. Default is 5.
-        /// Ex. If the limit is 5, up to 5 rotated files in addition to the original log file, will be kept. After reaching this limit,
-        /// the next time the log file is rotated, the oldest file will be deleted. If the limit is 0, no rotated files are kept,
-        /// i.e. when the log file is rotated, it is simply erased and starts over.
-        /// </summary>
-        public int RotateFileCountLimit { get; set; } = 5;
-
-        /// <summary>
         /// Creates a `FileLogHandler` that can write to the given file path.
         /// </summary>
         /// <param name="path">The text file to output messages to.</param>
-        /// <param name="logType">The <see cref="LogType"/> to use.</param>
-        /// <param name="shouldRotate">
-        /// If true, the file will be rotated when the size limit is reached. When a file is rotated, it has an index appended to it,
-        /// i.e. "log.txt" is renamed "log.1.txt", and a new "log.txt" is opened. If "log.1.txt" already exists, it is first renamed
-        /// "log.2.txt", and so on, rotating through log files. Once the file count limit is reached, the oldest file is deleted.
+        /// <param name="writeMode">The <see cref="WriteMode"/> to use.</param>
+        /// <param name="rotateFileSizeLimit">
+        /// The maximum file size in bytes before the file is rotated. Default is 10MB.
+        /// If set, the file will be rotated when the next log would exceed the limit. When a file is rotated, an index is appended to it,
+        /// ex. "log.txt" is renamed "log.1.txt", and a new "log.txt" is opened. If "log.1.txt" already exists, it is first renamed
+        /// "log.2.txt", and so on, rotating through log files.
         /// </param>
-        public FileLogHandler(string path, LogType logType = LogType.Append, bool shouldRotate = true)
+        /// <param name="rotateFileCountLimit">
+        /// The maximum number of rotated files to keep before deleting the oldest. Default is 5.
+        /// Ex. If the limit is 5, up to 5 rotated files in addition to the original log file, will be kept. After reaching this limit,
+        /// the next time the log file is rotated, the oldest file will be deleted. If the limit is 0, no rotated files are kept,
+        /// i.e. when the log file is rotated, it is simply erased and started over.
+        /// </summary>
+        public FileLogHandler(
+            string path,
+            WriteMode writeMode = WriteMode.Append,
+            long? rotateFileSizeLimit = 10L * 1024 * 1024,
+            int? rotateFileCountLimit = 5
+        )
         {
             m_Path = path;
-            m_LogType = logType;
-            m_ShouldRotate = shouldRotate;
+            m_WriteMode = writeMode;
+            m_RotateFileSizeLimit = rotateFileSizeLimit;
+            m_RotateFileCountLimit = rotateFileCountLimit;
+
+            if (m_RotateFileSizeLimit < 0)
+            {
+                throw new Exception($"Expected a positive integer for file size limit: {m_RotateFileSizeLimit}");
+            }
+            if (m_RotateFileCountLimit < 0)
+            {
+                throw new Exception($"Expected a positive integer for file count limit: {m_RotateFileCountLimit}");
+            }
 
             m_Directory = Path.GetDirectoryName(m_Path);
             m_FileName = Path.GetFileNameWithoutExtension(m_Path);
@@ -118,21 +132,18 @@ namespace Anvil.CSharp.Logging
 
             CreateWriter();
 
-            if (logType == LogType.Replace && shouldRotate)
+            if (m_WriteMode == WriteMode.Replace && m_RotateFileSizeLimit.HasValue)
             {
-                int index = 1;
-                string rotatedFilePath = GetRotatedFilePath(index);
-                while (File.Exists(rotatedFilePath))
+                foreach ((_, string filePath) in GetRotatedFiles())
                 {
-                    File.Delete(rotatedFilePath);
-                    rotatedFilePath = GetRotatedFilePath(++index);
+                    File.Delete(filePath);
                 }
             }
         }
 
         private void CreateWriter()
         {
-            m_Writer = new StreamWriter(m_Path, append: (m_LogType == LogType.Append))
+            m_Writer = new StreamWriter(m_Path, append: (m_WriteMode == WriteMode.Append))
             {
                 AutoFlush = true
             };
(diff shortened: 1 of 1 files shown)


Logging/LogHandler/FileLogHandler.cs
```diff
@@ -97,6 +158,56 @@ protected override void DisposeSelf()
                 );
 
             m_Writer.WriteLine($"{timestamp}{logLevel}{context}{message}");
+
+            if (m_Rotate && m_Writer.BaseStream.Length > RotateFileSizeLimit)
+            {
+                this.RotateFiles();
+            }
```
tjvezina, line 165: Do we need to worry about validating values like `RotateFileSizeLimit`, to ensure it's positive or even a reasonable minimum value? Or what you set is what you get?

Logging/LogHandler/FileLogHandler.cs
```diff
@@ -10,12 +9,30 @@ namespace Anvil.CSharp.Logging
...
+        /// The type of log files the handler should generate.
+        /// </summary>
+        public enum LogType
+        {
+            /// <summary>
+            /// New logs are appended to the existing log file.
+            /// </summary>
+            Append,
+            /// <summary>
+            /// A new log file is created each session, overwriting the previous one if it exists.
```
jkeon, line 22:
```suggestion
            /// A new log file is created, overwriting the previous one if it exists.
```

Super minor, but we should avoid mention of the session as the app layer will handle that

Logging/LogHandler/FileLogHandler.cs
```diff
@@ -10,12 +9,30 @@ namespace Anvil.CSharp.Logging
...
         public const string LOG_CONTEXT_CALLER_DERIVED_TYPE = "{0}";
         public const string LOG_CONTEXT_CALLER_FILE = "{1}";
         public const string LOG_CONTEXT_CALLER_METHOD = "{2}";
         public const string LOG_CONTEXT_CALLER_LINE = "{3}";
 
-        private readonly StreamWriter m_Writer;
+        private StreamWriter m_Writer;
+        private readonly string m_Path;
+        private readonly LogType m_LogType;
+        private readonly bool m_Rotate;
```
jkeon, line 35:
```suggestion
        private readonly bool m_ShouldRotate;
```
More clear?

Logging/LogHandler/FileLogHandler.cs
```diff
@@ -56,14 +73,58 @@ public class FileLogHandler : AbstractAnvilBase, ILogHandler
...
+        public long RotateFileSizeLimit { get; set; } = 10L * 1024 * 1024;
+
+        /// <summary>
+        /// The maximum number of rotated files to keep before deleting the oldest. Default is 5.
+        /// Ex. If the limit is 5, up to 5 rotated files in addition to the original log file, will be kept. After reaching this limit,
+        /// the next time the log file is rotated, the oldest file will be deleted. If the limit is 0, no rotated files are kept,
+        /// i.e. when the log file is rotated, it is simply erased and starts over.
+        /// </summary>
+        public int RotateFileCountLimit { get; set; } = 5;
+
```
jkeon, line 88: I'm not sure where but should we add some sort of `<remarks>` comment about how this work similar to the Unix rotate thing @mbaker3 was talking about in #108 ?

Logging/LogHandler/FileLogHandler.cs
```diff
@@ -56,14 +73,58 @@ public class FileLogHandler : AbstractAnvilBase, ILogHandler
...
         /// <param name="path">The text file to output messages to.</param>
-        /// <param name="append">If true, the file will have new logs appended to it, otherwise the file is replaced.</param>
-        public FileLogHandler(string path, bool append = true)
+        /// <param name="logType">The <see cref="LogType"/> to use.</param>
+        /// <param name="rotate">
+        /// If true, the file will be rotated when the size limit is reached. When a file is rotated, it has an index appended to it,
+        /// i.e. "log.txt" is renamed "log.1.txt", and a new "log.txt" is opened. If "log.1.txt" already exists, it is first renamed
+        /// "log.2.txt", and so on, rotating through log files. Once the file count limit is reached, the oldest file is deleted.
+        /// </param>
+        public FileLogHandler(string path, LogType logType = LogType.Append, bool rotate = true)
```
jkeon, line 99:
```suggestion
        public FileLogHandler(string path, LogType logType = LogType.Append, bool shouldRotate = true)
```

More clear?

Logging/LogHandler/FileLogHandler.cs
jkeon, line 35: Minor but since we're using an enum for `LogType` with two values, should keep the consistency and have an enum for `RotateBehaviour` instead of a bool?

Logging/LogHandler/FileLogHandler.cs
jkeon: I think we could `Debug.Assert` that it's at least greater than 0 if changing the type to `uint` is weird because of the API etc. 

If you set it to `int.MaxValue` though, that's on you. I don't think we should stop anyone because that probably means they don't care about diskspace and just wanted the "infinite" default which we don't have. It's doubtful they actually meant to rotate above `int.MaxValue`

Logging/LogHandler/FileLogHandler.cs
```diff
@@ -97,6 +158,56 @@ protected override void DisposeSelf()
...
+
+        private void RotateFiles()
+        {
+            m_Writer.Close();
+            m_Writer = null;
+
+            // Determine how many rotated log files already exist
+            int index = 1;
+            string rotatedFilePath = GetRotatedFilePath(index);
+            while (File.Exists(rotatedFilePath))
```
jkeon, line 176: Throughout this class there are a bunch of while loops and they all seem very safe so this is very much a 🚲 comment. 

I'm wondering if we should be explicitly safe and have them all be `for` loops since we know the start and know the upper limit of what could be. Not as clean to look at but guaranteed to never endless loop.

Logging/LogHandler/FileLogHandler.cs
```diff
@@ -97,6 +158,56 @@ protected override void DisposeSelf()
...
+            if (index == 0)
+            {
+                return m_Path;
+            }
+
+            string directory = Path.GetDirectoryName(m_Path);
+            string fileName = Path.GetFileNameWithoutExtension(m_Path);
+            string extension = Path.GetExtension(m_Path);
+            
+            return Path.Combine(directory, $"{fileName}.{index}{extension}");
```
jkeon, line 210: Minor since this wouldn't get called often, but we can cache these strings as readonly in the constructor.

jkeon (commented):
Looks awesome!

Pretty much all of my comments are just open ended questions/discussion. Feel free to ignore/disagree. 

I'll let @mbaker3 make the actual approve/request changes since he is much more involved with this feature.

Logging/LogHandler/FileLogHandler.cs
tjvezina: I was on the fence about this - `shouldRotate`? `useRotation`? `rotateFiles`?

tjvezina: My justification was that "Replace" isn't the obvious opposite of "Append", they're two distinct settings to me. Rotation though is just a switch - it's on or off. The enum would be `Rotate` and ... `DontRotate`?

Logging/LogHandler/FileLogHandler.cs
tjvezina: `int.MaxValue` is fine, but I meant if they set the file size limit to like... 5 (bytes) they're going to get a lot of log files. But maybe that's fine.

I'll just add an assert for `> 0`

Logging/LogHandler/FileLogHandler.cs
jkeon: my vote is `shouldRotate` but open to alternatives if someone feels strongly

Logging/LogHandler/FileLogHandler.cs
tjvezina: I ended up defining "file rotation" in the constructor under the `rotate` param, but let me know if there's a better way to document that.

Logging/LogHandler/FileLogHandler.cs
jkeon: Well come on, we're professionals, it would be `DO_NOT_ROTATE` obvi. /s

That's a fair point. I guess in my mind I had thought it would have been `Rotate` and `Replace` but that's confusing with `Replace` already being used in the `LogType`. 

Oh well, it was minor anyway, I'm cool with the bool

Logging/LogHandler/FileLogHandler.cs
jkeon: Yeah I think it's on them. The rotate limit will catch most of the damage from exploding amount of files and they'll just thrash the IO. Which hopefully they notice via profiling. I think it would be too difficult to assume a minimum otherwise.

Logging/LogHandler/FileLogHandler.cs
tjvezina: That first loop is the problem, since I don't yet know how many rotated log files exist. Even if `RotatedFileCountLimit` is 5, maybe you just lowered it from 10 and there are extra files I need to clean up. At best I could pick an arbitrary upper limit to stop at, but if that limit is reached, apparently there were just that many log files (unless there's a bug in the code I suppose).

The other two could easily be `for` loops though, I'll update those.

tjvezina: Now I'm unsure - I think the `for` loops are a bit harder to follow, and I'm not sure they're any less susceptible to infinite-loop-causing bugs. I'll keep whichever version you like more - thoughts?
```cs
// Determine how many rotated log files already exist
int index = 1;
string rotatedFilePath = GetRotatedFilePath(index);
while (File.Exists(rotatedFilePath))
{
    rotatedFilePath = GetRotatedFilePath(++index);
}
int rotatedFileCount = index - 1;

// Delete any files in excess of the file count limit
for (int i = rotatedFileCount; i >= RotateFileCountLimit; i--)
{
    File.Delete(GetRotatedFilePath(i));
    --rotatedFileCount;
}

// Rotate remaining files
for (int i = rotatedFileCount; i >= 0; i--)
{
    File.Move(GetRotatedFilePath(index), GetRotatedFilePath(index + 1));
}
```

Logging/LogHandler/FileLogHandler.cs
tjvezina: The best I can come up with is `Default`, which doesn't tell you anything. It's not that the logs would be replaced without rotation, it's that they would just grow indefinitely as logs come in, and the best way I can think to represent that is `shouldRotate: false`.

I will keep the bool until a stronger opinion appears!

Logging/LogHandler/FileLogHandler.cs
jkeon: Yeah in this case, I think the while loops are cleaner and I do think they are unlikely to cause issues. I guess I was more raising the question to the group of whether we should strive for safety in Anvil and where that grey muddy balance is. 

When I see a while loop, an alarm bell goes off but that shouldn't always mean that it can't be a while loop.

Logging/LogHandler/FileLogHandler.cs
```diff
@@ -94,9 +163,54 @@ protected override void DisposeSelf()
...
+            }
+
+            CreateWriter();
+        }
+
+        private string GetRotatedFilePath(int index)
+        {
+            Debug.Assert(index >= 0, $"Failed to get rotated file path, invalid index {index}");
+            return (index == 0 ? m_Path : Path.Combine(m_Directory, $"{m_FileName}.{index}{m_Extension}"));
         }
```
tjvezina, line 214: For what it's worth, I tested paths with no directory or extension, i.e. if you give the path "logfile" this will still work, and your rotated files will be "logfile.1" etc.

Also if you give a path under several folders that don't exist yet, ex. "some/new/folders/log.txt", the `Directory.CreateDirectory(...)` in the constructor creates each folder in the path before opening the file writer (this was technically a bug in the previous code).

mbaker3 (changes requested):
Looking good. Some minor suggestions, things to think about and a couple of easy fixes.

The one thing I want to propose is that we extract the rotation logic out into its own type. That way we can reuse it in other situations (crash logs, auto save, etc..) and would reduce this class back to one responsibility. 

A few of very half baked possible approaches that come to mind to inspire:
 - Write a `FileRotationMonitor` that would monitor a file and rotate it on a size threshold. The actual rotation and culling would be handled by a separate type (`FileRotator`?...could even be a static util)
 - Create a `RotatingStreamWriter` that handles the rotation monitoring. The stream instance is then passed into the `FileLogHandler`
 - Create an `IFIleLogGovernor` interface that the `FileLogHandler` would take in 1-n of and ask permission to write to the log file before actually writing. (Ex: `message = RotatingLogGovernor.PrepareToWrite(message)`)
    - Would also provide a hook to write a governor that would sanitize sensitive info from logs 🤷 

If that feels out of scope for this PR I'm totally fine merging this and extracting the logic later so long as we stub out an issue for the work.

Logging/LogHandler/FileLogHandler.cs
```diff
@@ -94,9 +163,54 @@ protected override void DisposeSelf()
...
+            if (m_ShouldRotate && m_Writer.BaseStream.Length > RotateFileSizeLimit)
+            {
+                this.RotateFiles();
+            }
+        }
+
+        private void RotateFiles()
+        {
+            Debug.Assert(RotateFileSizeLimit >= 0, $"Expected positive integer for file size limit, but got {RotateFileSizeLimit}");
+            Debug.Assert(RotateFileCountLimit >= 0, $"Expected positive integer for file count limit, but got {RotateFileCountLimit}");
```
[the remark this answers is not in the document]
tjvezina: `RotateFiles()` should be called quite infrequently, no? With the defaults, 10MB is about once every 10,000 logs.

I'm fine with wrapping in `#if DEBUG` either way - is `Debug.Assert(...)` not one of those that amounts to a no-op in release builds then?

Logging/LogHandler/FileLogHandler.cs
```diff
@@ -94,9 +163,54 @@ protected override void DisposeSelf()
...
+
+            m_Writer.Close();
+            m_Writer = null;
+
+            // Determine how many rotated log files already exist
+            int index = 1;
+            string rotatedFilePath = GetRotatedFilePath(index);
```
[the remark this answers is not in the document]
tjvezina: Good idea, I wasn't aware! I'll took at improving that.

Logging/LogHandler/FileLogHandler.cs
```diff
@@ -56,14 +78,61 @@ public class FileLogHandler : AbstractAnvilBase, ILogHandler
...
         /// <param name="path">The text file to output messages to.</param>
-        /// <param name="append">If true, the file will have new logs appended to it, otherwise the file is replaced.</param>
-        public FileLogHandler(string path, bool append = true)
+        /// <param name="logType">The <see cref="LogType"/> to use.</param>
+        /// <param name="shouldRotate">
+        /// If true, the file will be rotated when the size limit is reached. When a file is rotated, it has an index appended to it,
+        /// i.e. "log.txt" is renamed "log.1.txt", and a new "log.txt" is opened. If "log.1.txt" already exists, it is first renamed
+        /// "log.2.txt", and so on, rotating through log files. Once the file count limit is reached, the oldest file is deleted.
+        /// </param>
+        public FileLogHandler(string path, LogType logType = LogType.Append, bool shouldRotate = true)
```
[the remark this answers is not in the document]
tjvezina: I initially had it set up that way, but things got messy because I wanted different defaults for append vs. replace - that's changed, but it's still weird if you do something like pass a file count limit but no size limit. I guess that's fine, it will still work as instructed.

About the public setter, I was thinking about the syntax:
```
new FileLogHandler("log.txt")
{
    RotateFileCountLimit = 10;
}
```
...and figured changing those values after it's running was no stranger than changing `MinimumLevel` or `IncludeTimestamp`. Is there a benefit to the others being configurable post-instantiation, or could they all be readonly settings?

Logging/LogHandler/FileLogHandler.cs
tjvezina: Rebuilding this with `Directory.GetFiles()` rather than assuming any existing files will be sequential, so no one gets any loops!

Logging/LogHandler/FileLogHandler.cs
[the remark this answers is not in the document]
tjvezina: What should happen when there's a gap? If only 1 and 3 exist, should they become 2 and 3, or 2 and 4?

I'm guessing the former, i.e. don't necessarily rotate every log file, only do so when attempting to rotate a newer file into its place.

Logging/LogHandler/FileLogHandler.cs
tjvezina: Test cases are now summarized in PR description

Logging/LogHandler/FileLogHandler.cs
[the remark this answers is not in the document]
tjvezina: I'll leave the open question, but I've made this change anyway.

Logging/LogHandler/FileLogHandler.cs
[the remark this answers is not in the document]
mbaker3: I'd say cull anything after continuity is broken. Otherwise could lead to some confusing situations where you think two adjacent log files are connected through a session but really aren't.

Of course, you could look at the date/time on the message but it would be an easy oversight.

Logging/LogHandler/FileLogHandler.cs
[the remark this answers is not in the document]
mbaker3: > I initially had it set up that way, but things got messy because I wanted different defaults for append vs. replace - that's changed, but it's still weird if you do something like pass a file count limit but no size limit. I guess that's fine, it will still work as instructed.

It would be an odd configuration but wouldn't break anything. I think that's ok.

> About the public setter, I was thinking about the syntax:
> ...

Yea, I appreciate the syntax for sure. In later versions of C# we'll be able to mark the setters with `init` and this will work great.

I suppose the difference with `IncludeTimestamp` and `MinimumLevel` is that they're not risky runtime changes. Your logs will just come out looking different after the change.

I think there's definitely a case for allowing `MinimumLevel` to be changed at runtime. The need to change `IncludeTimestamp` is much less likely.

Logging/LogHandler/FileLogHandler.cs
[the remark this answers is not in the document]
mbaker3: Oof, you're right sorry. For some reason I thought `RotateFiles()` was called on every `HandleLog()` call. Missed the `if` check 😬

Logging/LogHandler/FileLogHandler.cs
```diff
@@ -118,21 +132,18 @@ public FileLogHandler(string path, LogType logType = LogType.Append, bool should
...
+            if (m_WriteMode == WriteMode.Replace && m_RotateFileSizeLimit.HasValue)
             {
-                int index = 1;
-                string rotatedFilePath = GetRotatedFilePath(index);
-                while (File.Exists(rotatedFilePath))
+                foreach ((_, string filePath) in GetRotatedFiles())
                 {
-                    File.Delete(rotatedFilePath);
-                    rotatedFilePath = GetRotatedFilePath(++index);
+                    File.Delete(filePath);
```
mbaker3, line 139: Should we log what is being deleted?

Logging/LogHandler/FileLogHandler.cs
```diff
@@ -165,52 +176,107 @@ protected override void DisposeSelf()
...
             {
-                this.RotateFiles();
+                int logLength = Encoding.Unicode.GetByteCount(log);
+                long fileLength = m_Writer.BaseStream.Length;
+
+                if ((fileLength + logLength) > m_RotateFileSizeLimit)
+                {
+                    if (logLength > m_RotateFileSizeLimit)
+                    {
+                        throw new Exception("Message size exceeds file size limit, logging failed " +
```
mbaker3, line 190: Apologies if I asked for an exception here. We should probably just log an error or cull the message and append some `{culledMessage}...(message longer than file size limit)` type message.

No sense in crashing the application because a log that's too long slipped in. Thinking of cases where we log serialized JSON state and it's fine until suddenly it's too big some day and isn't ok.

Same goes for crazy stacks.

Logging/LogHandler/FileLogHandler.cs
```diff
@@ -165,52 +176,107 @@ protected override void DisposeSelf()
...
             }
-            --index;
 
-            // Delete any files in excess of the file count limit
-            while (index >= RotateFileCountLimit)
+            RotateRecursive(0);
+
+            CreateWriter();
+
+            void RotateRecursive(int index)
```
mbaker3, line 236: Tell me you've been writing a lot of TS/JS without telling me 😄

Logging/LogHandler/FileLogHandler.cs
```diff
@@ -79,33 +84,42 @@ public enum LogType
...
-        /// <param name="shouldRotate">
-        /// If true, the file will be rotated when the size limit is reached. When a file is rotated, it has an index appended to it,
-        /// i.e. "log.txt" is renamed "log.1.txt", and a new "log.txt" is opened. If "log.1.txt" already exists, it is first renamed
-        /// "log.2.txt", and so on, rotating through log files. Once the file count limit is reached, the oldest file is deleted.
-        /// </param>
-        public FileLogHandler(string path, LogType logType = LogType.Append, bool shouldRotate = true)
+        public FileLogHandler(
+            string path,
+            WriteMode writeMode = WriteMode.Append,
+            long? rotateFileSizeLimit = 10L * 1024 * 1024,
```
mbaker3, line 106: Add 10mb comment or move to a constant and put the comment there.

Logging/LogHandler/FileLogHandler.cs
tjvezina: I meant to wrap all the exceptions in `#if DEBUG` but yeah, I can have it trim the log and append a message about it being truncated. Hard to log a warning though, as that would just be pushed to the next log file, and may start an infinite truncation warning loop if you set a really small file size limit by accident.

Logging/LogHandler/FileLogHandler.cs
tjvezina: I don't think so, if you've explicitly chosen `WriteMode: Replace` then you expect the files (rotated or not) to be replaced. We didn't send a log pre-rotation when replacing an existing log.

Logging/LogHandler/FileLogHandler.cs
tjvezina: I don't see it, what did I do? 😛

Logging/LogHandler/FileLogHandler.cs
tjvezina: The `<param>` description above says the default is 10MB, or did you want another comment inline here?

Logging/LogHandler/FileLogHandler.cs
mbaker3: inline functions. Much less frequently used in C# unless they're lambdas

tjvezina: Oh, I used local functions long before I learned JS - any time I want to encapsulate logic relevant only to the current function, for use as a callback, LINQ predicate, recursion, etc.

Logging/LogHandler/FileLogHandler.cs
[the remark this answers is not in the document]
tjvezina: I could do that too - I can still use `Directory.GetFiles()` to pull a list of rotated files, but instead of handling them all I would just step through them sequentially starting at 1, and as soon as I reach a gap, delete all remaining files? Then I could remove all the custom handling I've currently got, which attempts to preserve any files it finds.

Logging/LogHandler/FileLogHandler.cs
tjvezina: The problem with truncating a message to a certain size in bytes is that char count != byte count. Maybe you have lots of 2-byte unicode characters in your log. I can think of two ways to handle this:

1. Convert to byte array, truncate, convert to string
    - this risks cutting a unicode char in half and "corrupting" the last visible character in the log.
2. Truncate the string to `targetByteCount` chars to start, then continue trimming one char at a time and checking `Encoding.Unicode.GetByteCount(log)` until it's under the limit
    - gross but functional, and very very rarely used, unless you've set your file size limit to 10KB and are logging massive JSON files or something

Logging/LogHandler/FileLogHandler.cs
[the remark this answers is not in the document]
mbaker3: > I would just step through them sequentially starting at 1, and as soon as I reach a gap, delete all remaining files?

Exactly what I was thinking.

Logging/LogHandler/FileLogHandler.cs
mbaker3: For option 2, would it be better to
 - Go backward one character at a time
 - Measure the byte size of that one character
 - Repeat until your threshold is met?

**Option 3** 
On the other hand, accuracy isn't the most important thing in the world and most characters are 1-byte in UTF-8 (the ASCII range) so you could assume that 1 character == 1 byte and worst case you'd remove 4x the number of characters you needed to. 🤷 

I'm goo with either Option 2 or 3. Just throw a comment in there describing what we're doing.

Life would be easy if we just ASCII encoded but that could be restrictive...and would support emojis.

Logging/LogHandler/FileLogHandler.cs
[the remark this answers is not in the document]
tjvezina: Done.

Logging/LogHandler/FileLogHandler.cs
tjvezina: We definitely need unicode support for emojis... and non-English names I guess. 😉

Things got hairy here, but I've pushed a fairly reasonable solution:
- Each log goes through `m_Writer.Encoding.GetMaxByteCount(...)` to determine if we're anywhere near the file size limit
  - _This is a cheap check, basically assuming every char is the maximum byte count for the current system's encoding_
- If the next log may go over the limit, call the more expensive `m_Writer.Encoding.GetByteCount()`
- Remove at least enough chars to guarantee this log will be under the limit (including appended truncation message)
  - If the culled chars were all ASCII, the file will be exactly the file size limit
  - If any culled chars were two-byte unicode, we're just trimming a bit extra and coming in a little under the limit

Tested by creating a `FileLogHandler` with a size limit of 1024 bytes, and logging a 1000-char string with some unicode chars in it. With the addition of the log context and timestamp (and a system-specific newline added by the writer when you call `WriteLine(...)`) this goes well over the limit, is truncated, and has a truncation message appended, resulting in a file exactly 1KB (1024 B) in size on my PC.

tjvezina:
If the latest changes are acceptable (or after any last rounds of minor adjustments), feel free to merge this PR - I think it makes sense to later open a new PR for the separation of file rotation logic from `FileLogHandler`, not least because of how busy this one has already gotten.

mbaker3 (changes requested):
This looks great. One comment about the constructor argument assertions but otherwise I think we're good.

Thanks for putting this together. This is some very nice functionality!

mbaker3 (approved):

PR #113, opened 2022-10-06
Logging polish

tjvezina:
1. If log caller file/name are unknown, send "unknown" to log handlers instead of null
2. Changed log context format from `(FILE|METHOD:LINE)` to `(FILE:LINE|METHOD)`

Note: The default context string was initially going to be "\<unknown>" (with angle brackets), but each handler is expecting a file path and runs the given string through `Path.GetFileNameWithoutExtension()`, which throws an exception if given a string with invalid path characters. There are other options like "[unknown]" but for now I used no additional characters, as the text already stands out by being lowercase, beside the PascalCase of C# types/files.

If for any reason a log's context fails to resolve, the context prepended to the message says `(|:0) message`

If for any reason a log's context fails to resolve, the context will now say `(unknown:0|unknown) message`

Improved log readability

 - None

 - [ ] Yes
 - [X] No

### Related
- https://github.com/decline-cookies/anvil-unity-core/pull/80
- https://github.com/decline-cookies/anvil-unity-core/issues/79

### DIFF
--- a/Logging/.CSProject/Log.cs
+++ b/Logging/.CSProject/Log.cs
@@ -11,6 +11,8 @@ namespace Anvil.CSharp.Logging
     /// </summary>
     public static class Log
     {
+        private const string UNKNOWN_CONTEXT = "unknown";
+
         private static readonly HashSet<ILogHandler> s_AdditionalHandlerList = new HashSet<ILogHandler>();
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


jkeon (approved):
Thumbs up emoji

mbaker3 (approved):
Great change 🚢 🚢 🚢 🚢

PR #114, opened 2022-10-06
Hotfix id static reset

mbaker3:
Add a method to allow the static state of `ID` to be reset.
This is useful in run times that do not perform a domain reload between run instances (Ex Unity Editor).

#### Tag Alongs
 - Rename `ID.IDLimitWarningEventArgs` to `ID.LimitWarningEventArgs`. The outer class already qualifies the ID context of the type.

Instances that do not correctly remove their listener from `ID.OnIDLimitWarning` when execution stops will be notified of event triggers for runtimes that do not perform a domain reload.

Runtimes that do not perform a domain reload can now request `ID` to reset its static state.

None

 - None

 - [x] Yes - Rename any references to `ID.IDLimitWarningEventArgs`
 - [ ] No

### DIFF
--- a/Data/ID/ID.cs
+++ b/Data/ID/ID.cs
@@ -1,6 +1,5 @@
 using Anvil.CSharp.Logging;
 using System;
-using System.ComponentModel;
 
 namespace Anvil.CSharp.Data
 {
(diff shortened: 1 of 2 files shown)


Data/ID/ID.cs
```diff
@@ -1,5 +1,6 @@
 using Anvil.CSharp.Logging;
 using System;
+using System.ComponentModel;
```
jkeon, line 3: accidental import?

jkeon (changes requested):
Looks good, one minor accident?

Data/ID/ID.cs
mbaker3: yep, fixed. Thanks.

jkeon (approved):

PR #115, opened 2022-10-06
Logging polish (part 2)

tjvezina:
1. Omit line number from log context when zero
2. Corrected outdated documentation comment

Logs from managed code have line number zero in the context, ex: `(UnityLogListener:0|.ctor) message`

Logs from managed code now omit the line number, ex: `(UnityLogListener|.ctor) message`

Note: The same functionality does not apply to `FileLogHandler` as it allows the user to define a log context format string, making it confusing if part of that format is silently omitted in some cases. It would be nearly impossible to cleanly omit the line number in custom format strings anyway.

 - None

 - [ ] Yes 
 - [X] No

### Related
- https://github.com/decline-cookies/anvil-unity-core/pull/82

### DIFF
--- a/Logging/.CSProject/LogHandler/ConsoleLogHandler.cs
+++ b/Logging/.CSProject/LogHandler/ConsoleLogHandler.cs
@@ -23,11 +23,11 @@ namespace Anvil.CSharp.Logging
 
             if (callerLine > 0)
             {
-               message = $"({callerFile}:{callerLine}|{callerName}) {message}"
+               message = $"({callerFile}:{callerLine}|{callerName}) {message}";
             }
             else
             {
-               message = $"({callerFile}|{callerName}) {message}"
+               message = $"({callerFile}|{callerName}) {message}";
             }
 
             switch (level)
--- a/Logging/.CSProject/LogHandler/ConsoleLogHandler.cs
+++ b/Logging/.CSProject/LogHandler/ConsoleLogHandler.cs
@@ -21,9 +21,14 @@ namespace Anvil.CSharp.Logging
         {
             string callerFile = Path.GetFileNameWithoutExtension(callerPath);
 
-            string context = (callerLine > 0 ? $"{callerFile}:{callerLine}|{callerName}" : $"{callerFile}|{callerName}");
-
-            message = $"({context}) {message}";
+            if (callerLine > 0)
+            {
+               message = $"({callerFile}:{callerLine}|{callerName}) {message}"
+            }
+            else
+            {
+               message = $"({callerFile}|{callerName}) {message}"
+            }
 
             switch (level)
             {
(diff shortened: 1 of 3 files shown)


jkeon (dismissed):
Looks good. 

Why would the line number be 0 ever though? Maybe I'm not fully understanding the case when it is 0?

tjvezina:
It's just the default value returned by `StackFrame.GetFileLineNumber()` - so if the log context fails to resolve you'd see `(unknown|unknown)`, or more likely if the log comes from a DLL or other managed code, there is no file/line to point to, so we fall back on showing `DeclaringType.Name` and omit the line number, ex `(UnityLogListener|.ctor)`.

Logging/.CSProject/LogHandler/ConsoleLogHandler.cs
```diff
@@ -19,7 +19,11 @@ public class ConsoleLogHandler : ILogHandler
             string callerName,
             int callerLine)
         {
-            message = $"({Path.GetFileNameWithoutExtension(callerPath)}:{callerLine}|{callerName}) {message}";
+            string callerFile = Path.GetFileNameWithoutExtension(callerPath);
+
+            string context = (callerLine > 0 ? $"{callerFile}:{callerLine}|{callerName}" : $"{callerFile}|{callerName}");
+
+            message = $"({context}) {message}";
```
mbaker3, line 26: Same as comment here. Remove extra string allocation
https://github.com/decline-cookies/anvil-unity-core/pull/82#discussion_r990223627

mbaker3 (changes requested):
> Note: The same functionality does not apply to FileLogHandler as it allows the user to define a log context format string, making it confusing if part of that format is silently omitted in some cases. It would be nearly impossible to cleanly omit the line number in custom format strings anyway.

Yea that's a shame because it's the direction all of the handlers will move eventually.
When we do take on that effort maybe we just include two different versions of the caller line token.
`LOG_CONTEXT_CALLER_LINE_ALWAYS` and `LOG_CONTEXT_CALLER_LINE` where one would only show conditionally, maybe with one way to associate the separator with the conditional version so they could both be omitted effectively when unknown.

I suppose the other option is just to replace line number with `?` or `NaN` or something when it's not known.

mbaker3 (dismissed):

Logging/.CSProject/LogHandler/ConsoleLogHandler.cs
```diff
@@ -19,7 +19,16 @@ public class ConsoleLogHandler : ILogHandler
...
+            string callerFile = Path.GetFileNameWithoutExtension(callerPath);
+
+            if (callerLine > 0)
+            {
+               message = $"({callerFile}:{callerLine}|{callerName}) {message}"
+            }
+            else
+            {
+               message = $"({callerFile}|{callerName}) {message}"
+            }
```
mbaker3, line 31:
```suggestion
            if (callerLine > 0)
            {
               message = $"({callerFile}:{callerLine}|{callerName}) {message}";
            }
            else
            {
               message = $"({callerFile}|{callerName}) {message}";
            }
```

mbaker3 (approved):

jkeon:
Ahh gotcha. Ok cool thanks

PR #117, opened 2022-10-07
NUnit test framework setup

tjvezina:
A test project for writing and running NUnit tests.

- A lack of unit testing to guard against regression

 - None

 - [ ] Yes 
 - [X] No

### Related
- https://github.com/decline-cookies/anvil-unity-core/pull/84

---

### Running Tests - Pure C#/.NET Projects
The `anvil-csharp-tests.csproj` should be added to your project solution, where you can then write and run tests. Depending on your IDE, tests are often automatically detected and appear in the IDE's testing interface.

#### Example: Visual Studio - Test Explorer
image1

#### Example: VS Code - [.NET Core Test Explorer](https://marketplace.visualstudio.com/items?itemName=formulahendry.dotnet-test-explorer)
image3

_(I don't have access to Rider, but others are welcome to explore its NUnit support in pure C# projects)._

### Running Tests - Unity Projects
The `anvil-csharp-tests.asmdef` is picked up by Unity and used to package all files in the `Tests` folder for running. The .csproj used in pure C# projects is ignored.

image2

### NUnit
[NUnit Reference](https://docs.nunit.org/articles/nunit/writing-tests/assertions/assertions.html)

Writing tests is as easy as adding a `[Test]` attribute to a function, and using NUnit's `Assert.That(...)` system to test functionality. 

Here's an example script:
```cs
using NUnit.Framework;
using Anvil.CSharp.Logging;

namespace Anvil.CSharp.Tests
{
    public static class LoggerTests
    {
        [Test]
        public static void TestStaticLogger()
        {
            Logger logger = Log.GetStaticLogger(typeof(LoggerTests));

            Assert.That(logger.DerivedTypeName, Is.EqualTo(nameof(LoggerTests)));
        }
    }
}
```

### DIFF
--- a/Tests/anvil-csharp-tests.asmdef
+++ b/Tests/anvil-csharp-tests.asmdef
@@ -1,0 +1,16 @@
+{
+    "name": "anvil-csharp-tests",
+    "rootNamespace": "Anvil.CSharp.Tests",
+    "references": [],
+    "includePlatforms": [
+        "Editor"
+    ],
+    "excludePlatforms": [],
+    "allowUnsafeCode": false,
+    "overrideReferences": false,
+    "precompiledReferences": [],
+    "autoReferenced": true,
+    "defineConstraints": [],
+    "versionDefines": [],
+    "noEngineReferences": false
+}
--- a/Tests/anvil-csharp-tests.csproj
+++ b/Tests/anvil-csharp-tests.csproj
@@ -1,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <RootNamespace>Anvil.CSharp.Tests</RootNamespace>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.3.2" />
+    <PackageReference Include="NUnit" Version="3.13.3" />
+    <PackageReference Include="NUnit3TestAdapter" Version="4.2.1" />
+    <PackageReference Include="NUnit.Analyzers" Version="3.3.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\anvil-csharp-core.csproj" />
+  </ItemGroup>
+
+</Project>
--- a/anvil-csharp-core.csproj
+++ b/anvil-csharp-core.csproj
@@ -1,10 +1,16 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <RootNamespace>anvil_csharp_core</RootNamespace>
+    <RootNamespace>Anvil.CSharp</RootNamespace>
     <LangVersion>8</LangVersion>
     <TargetFrameworks>netcoreapp3.1;netstandard2.1</TargetFrameworks>
   </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Remove="Tests\**" />
+    <EmbeddedResource Remove="Tests\**" />
+    <None Remove="Tests\**" />
+  </ItemGroup>
 
   <ItemGroup>
     <None Remove="Logging\anvil-csharp-logging.dll" />


Tests/anvil-csharp-tests.asmdef
```diff
@@ -0,0 +1,16 @@
...
+        "Editor"
+    ],
+    "excludePlatforms": [],
+    "allowUnsafeCode": false,
+    "overrideReferences": false,
+    "precompiledReferences": [],
+    "autoReferenced": true,
+    "defineConstraints": [],
+    "versionDefines": [],
+    "noEngineReferences": false
```
jkeon, line 15: We probably want to exclude Unity engine references for any pure c-sharp libs

jkeon (dismissed):
Looks good, thanks for doing this.

One minor change.

Tests/anvil-csharp-tests.asmdef
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "anvil-csharp-tests",
+    "rootNamespace": "Anvil.CSharp.Tests",
+    "references": [],
```
jkeon, line 4: Probably want to reference anvil?

`anvil-csharp-core`

tjvezina: Same answer as https://github.com/decline-cookies/anvil-unity-core/pull/84 - the .asmdef automatically references other assemblies as needed by default. If explicit references are preferred, I will add them.

Tests/anvil-csharp-tests.asmdef
tjvezina: There's already an `anvil-csharp-core.asmdef` in the root of the repo for the same purpose - it's meaningless and harmless in non-Unity projects, and prevents there from being setup steps before Anvil is functional in Unity projects.

jkeon: Ok sound good!

Tests/anvil-csharp-tests.csproj
```diff
@@ -0,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
```
mbaker3, line 4: Should we match the main `anvil-csharp-core` csproj?
`netcoreapp3.1;netstandard2.1`

tjvezina: Oh right - I tried that, but it yelled because NUnit uses cutting edge .NET features, so I left it as `net6.0` since this is just a unit test project, and that was the default when I created it.

mbaker3: Seems like at least .NET Core 3.1 should be supported based on the release notes. 
https://docs.nunit.org/articles/nunit/release-notes/framework.html

Maybe it was just .NET Standard that was the issue?
https://docs.nunit.org/articles/nunit/getting-started/dotnet-core-and-dotnet-standard.html#why-cant-my-tests-target-net-standard

In any case, we should target the correct framework in case there are breaking, behavioural changes (unlikely) or performance changes.

The standard build actions should catch situations where a developer uses a feature in a newer version of .NET but it would be confusing so see the tests passing while the build action fails.

tjvezina: Wrapping up an offline discussion:

The primary concern is that, IDEs can sometimes allow language features to be used and committed that aren't actually supported in a given project's target framework, after which the PR build step should fail, but we want to make sure the unit test step (when it exists) doesn't confusingly succeed.

This isn't a problem, because while the anvil-csharp-tests project targets .NET 6.0, it still compiles and references anvil-csharp-core separately, failing if any unsupported features are used. It is impossible for builds to fail and unit tests to pass.

---

In an attempt to generate a bit of empirical evidence, I ran `dotnet test` to confirm that each project is compiled independently:
```
  anvil-csharp-logging -> .\Logging\anvil-csharp-logging.dll
  anvil-csharp-core -> .\bin\Debug\netcoreapp3.1\anvil-csharp-core.dll
  anvil-csharp-tests -> .\Tests\bin\Debug\net6.0\anvil-csharp-tests.dll
```

...after which I added a `global using` (a .NET 6.0 / C# 10.0 feature) to a unit test file with no issue, but adding one to a random anvil-csharp-core file caused the build to fail as expected:
```
Core\AbstractAnvilBase.cs(2,1): error CS8400: Feature 'global using directive' is not available in C# 8.0. Please use language version 10.0 or greater. [anvil-csharp-core.csproj]
```

mbaker3: Perfect, thanks for clearing that up for me and discussing. False alarm!

mbaker3 (approved):

Issue #103, opened 2022-08-16
Convert CircleCI build steps to GitHub Actions


mbaker3:
Current auto build validation runs using CircleCI. Convert to use Github Actions.

Remember to remove the shield from `README.md`. See if there's a Github Actions equivalent.
Looks like there's at least a shields.io version: https://shields.io/category/build

PR #119, opened 2022-10-09
GitHub Actions setup

tjvezina:
Replaced CircleCI config with GitHub Actions workflows. The result is the same, but builds are run and can be reviewed right here in GitHub.

Rather than separate runs for anvil-csharp-core and anvil-csharp-logging, both are built in a single run but as separate steps, which are easy to differentiate in GitHub's interface. There are still separate workflows for `dotnet` and `msbuild` builds, and both debug and release configurations are tested.

_Note: The GitHub Actions free plan includes 2000 min/month of operation._

PRs (and `main` pushes?) trigger CircleCI builds.

PRs and `main` pushes trigger GitHub Actions workflows.

- #103 - CI builds are a bit more accessible, readable, and configurable

 - None

 - [ ] Yes
 - [X] No

---

### Examples

- https://github.com/tjvezina/anvil-csharp-core/pull/1
- https://github.com/tjvezina/anvil-csharp-core/pull/4
- https://github.com/tjvezina/anvil-csharp-core/pull/3

In the event of a failure, GitHub makes it easy to [see which build step failed](https://github.com/tjvezina/anvil-csharp-core/actions/runs/3215101031/jobs/5255959453):

image


README.md
```diff
@@ -1,5 +1,7 @@
 ![GitHub](https://img.shields.io/github/license/decline-cookies/anvil-csharp-core)
-[![CircleCI](https://circleci.com/gh/decline-cookies/anvil-csharp-core/tree/main.svg?style=shield&circle-token=75e18a1354e8eb105cb7601a6f640b23e6975da4)](https://circleci.com/gh/decline-cookies/anvil-csharp-core/tree/main)
+![GitHub Workflow Status](https://img.shields.io/github/workflow/status/decline-cookies/anvil-csharp-core/.NET%20Build?label=.NET%20Build)
+![GitHub Workflow Status](https://img.shields.io/github/workflow/status/decline-cookies/anvil-csharp-core/MSBuild?label=MSBuild)
+
```
tjvezina, line 4: I've never worked with shields, but hopefully this is right? Not sure how to test until the workflow is actually running on the main branch.

tjvezina:
Hmm, not sure what happened to the MSBuild workflow on this PR, it seems to have quit because it was looking for the `github-actions-setup` branch in this repo instead of my fork. 🤔

README.md
tjvezina: Never mind, thought shields.io was the definitive source of project status shields, but Mike pointed me to the official GitHub workflow badges, which are now set up.

jkeon (dismissed):
Seems good to me

tjvezina:
Presumably CircleCI has to be disabled separately if this PR is accepted, and I have my only idea for the broken MSBuild (see [anvil-csharp-core/actions](https://github.com/decline-cookies/anvil-csharp-core/actions)) is that the config is being run from my fork I guess, so maybe after merging it will find branches correctly? I can help sort that out next, but at least .NET Build works.

.github/workflows/dotnet-build.yml
```diff
@@ -0,0 +1,29 @@
+name: .NET Build
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+jobs:
+  build:
```
mbaker3, line 7: We need to figure out a way to extract a more useful name out of the action. When selecting the status checks in the branch protection I see the following:
image

That's not super helpful for knowing what I'll be selecting. 

See how the old Circle-ci used a different name (Ex: `build-mono`)

.github/workflows/msbuild.yml
```diff
@@ -0,0 +1,31 @@
+name: MSBuild
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+jobs:
+  build:
+    runs-on: windows-latest
```
mbaker3, line 8: The CircleCI config ran `dotnet` in a windows environment and `msbuild` on linux (debian).

Any reason for switching?

Really we just want to make sure anvil gets compiled once with .NET and once with Mono. It just happens the docker image for Mono was based off Linux.
https://github.com/mono/docker/blob/9293c0cddf31a3dc829fccc6f8e1eb507a91cd34/6.12.0.182/slim/Dockerfile

.github/workflows/msbuild.yml
```diff
@@ -0,0 +1,31 @@
+name: MSBuild
```
mbaker3, line 1: Can we name these based on the objective rather than the tool used?
Ex: Mono

tjvezina: I really wanted to match ".NET Build" but "MSBuild Build" sounded weird. I can use whatever name you like though - what's the connection between msbuild and mono?

.github/workflows/msbuild.yml
tjvezina: Oh strange, I just noticed that every workflow example used `ubuntu-latest` so I did the same, but I thought msbuild needed windows - I can change it so both use `ubuntu-latest`? Or whatever arrangement you prefer.

.github/workflows/dotnet-build.yml
tjvezina: Hmm, okay - that's just the job name within the workflow (`build`) and a list of "matrix parameters" used in generating job variations (in this case, build configurations). I can change the job name to anything though, like `dotnet-build` and `mono-build`?

.github/workflows/msbuild.yml
mbaker3: hmmm, leave them different I guess

.github/workflows/msbuild.yml
mbaker3: When I set it up I believed `msbuild` was building with Mono but that's not necessarily the case. I was using the mono docker image so it likely was compiling with mono but it looks like the correct approach would be to compile using `csc`

https://www.mono-project.com/docs/getting-started/mono-basics/

I'd recommend renaming this workflow to `Mono Build`

We really just want to make sure anvil compiles with .NET and Mono since there are sometimes subtle differences. As we build out unit tests this will become more important.

mbaker3:
I've removed CircleCI from Decline Cookies and this repo. The checks still show circleCI failing...although with no workflows. I think next time a commit is pushed it should be removed from the checks.

.github/workflows/dotnet-build.yml
mbaker3: Sounds good. Maybe reverse the terms like we had before? My thought is that the actions build, test, etc.. are more likely to fail together vs specific compilers failing. So the failures will tend to show grouped in the UI? (IE: Failures next to one another)

Also, more likely that a specific action isn't supported by all platforms so it's better to group by action?
Ex:
 - build-dotnet
 - build-mono
 - test-dotnet
 - test-dotnet
 - profile-dotnet

In this example I'm assuming there is no headless profiling support from Mono.

tjvezina: The list of tests on a PR seems to have arbitrary order, I guess whichever ones start running first since they're all done in parallel, but maybe I'm not seeing a context where they're sorted.

Either way, I have nothing against reversing the naming, so `build-dotnet`/`build-mono` it is!

mbaker3 (approved):

Issue #112, opened 2022-10-03
Debug - Helper static function for nice type names


jkeon:
I've found I'm writing this all over the place:

```
m_TypeString = type.IsGenericType
                ? $"{type.Name[..^2]}<{type.GenericTypeArguments[0].Name}>"
                : type.Name;
```

I want to nicely output logs with the type of object I'm working with but I don't want to see the fully qualified class name all the time as it's tricky to read. 

Instead of `Anvil.Unity.DOTS.Entities.ProxyDataStream\`, `DeclineCookies.TwistedDots.Game.Simulation.TimerInstance`

I'd like to see it like you would in code with `ProxyDataStream<TimerInstance>`

The above example assumes only one generic type so we should have a general solution that adds the comma separators for N types. 

`MainType<GType1, GType2, GType3, GTypeN>`

mbaker3:
Is this for incorporating a type other than the calling type name into a log string? Curious if `Logger` currently captures a good name for generics.

jkeon:
I have a class that's a `EntityProxyDataStream<TimerInstance>`. 

When i look at the `Type` for that its:

`Anvil.Unity.DOTS.Entities.Tasks.EntityProxyDataStream``1[DeclineCookies.TwistedDots.Game.Simulation.TimerInstance]`

Which is a mouthful and makes the logs really polluted. 

But if I just do `Type.Name` I get: 

`EntityProxyDataStream``1`

What I really want is how I would see it in the code as `EntityProxyDataStream<TimerInstance>` and this accomplishes that.

The logs are clean and easier to parse.

mbaker3:
Is that what it looks like in the context that logger generates as well though? Ex: what's in the {myType} part of the log message in this example when logging?
`({myType}|MyMethod:0) Some message`

Just trying to figure out if the solution needs to be applied to logger as well or if logger provides a better result that we can steal the approach from. Having to resolve the and format the type string at runtime isn't ideal.

jkeon:
Ah sorry, I see, I hadn't checked. All my "logs" were being thrown as exceptions so they weren't going through Logger.

Just did a quick test and I get this:

`(EntityProxyDataStream|.ctor:33) Test`

So it does appear that some work needs to be done in Logger to accommodate the generics part since that's not being included.

mbaker3:
ok, good to know. Will have to apply this fix to `ILogger` implementations as well then.

[...]

PR #120, opened 2022-10-11
Type extension to get human-readable generic type names

tjvezina:
A type extension that handles generic types (and nullables) to generate a short human-readable type name, useful in logging.

_Note 1: It is not possible to format primitive types as they appear in code (i.e. `int` vs `Int32`) without access to `CSharpCodeProvider`, which Unity seems not to include, and would be expensive anyway. It seems sufficient to show the underlying type name for primitives anyway._

_Note 2: As mentioned in https://github.com/decline-cookies/anvil-csharp-core/issues/112, this change is inaccessible within the logging library, which therefore uses a static helper function that duplicates the type extension. A comment above said helper explains this._

Calling `typeof(myObj).ToString()` with generics gives results like this:
```
Anvil.Unity.DOTS.Entities.Tasks.EntityProxyDataStream`1[DeclineCookies.TwistedDots.Game.Simulation.TimerInstance]
```

Calling `typeof(myObj).GetReadableName()` will give results like this:
```
EntityProxyDataStream<TimerInstance>
```

#112 - Type names are easier to read in logs, without the need for a custom formatter every time

 - None

 - [ ] Yes
 - [X] No

---

### Examples

Here is the output from a variety of example types after calling `GetReadableName()`:
```
TypeA
GenTypeA<String>
GenTypeA<Int32>
GenTypeA<Int32?>
GenTypeA<TypeA>
GenTypeA<List<String>>
GenTypeA<GenTypeA<GenTypeA<Int32>>>
GenTypeB<String, Int32>
GenTypeC<Int32, Int32, Int32, Int32, Int32, Int32, Int32, Int32, Int32, Int32>
NestedTypeA
NestedGenTypeA<String>
NestedGenTypeA<Int32>
NestedGenTypeA<Int32?>
NestedGenTypeA<TypeA>
NestedGenTypeA<NestedTypeA>
NestedGenTypeA<List<String>>
NestedGenTypeA<NestedGenTypeA<NestedGenTypeA<Int32>>>
NestedGenTypeB<String, Int32>
NestedGenTypeC<Int32, Int32, Int32, Int32, Int32, Int32, Int32, Int32, Int32, Int32>
```

### DIFF
--- a/Logging/.CSProject/Logger/Logger.cs
+++ b/Logging/.CSProject/Logger/Logger.cs
@@ -35,9 +35,7 @@ namespace Anvil.CSharp.Logging
             int removeCount = 1 + (type.GenericTypeArguments.Length < 10 ? 1 : 2);
             string name = type.Name[..^removeCount];
 
-            string genericTypeNames = type.GenericTypeArguments
-                .Select(GetReadableName)
-                .Aggregate((a, b) => $"{a}, {b}");
+            string genericTypeNames = string.Join(", ", type.GenericTypeArguments.Select(GetReadableName));
 
             return $"{name}<{genericTypeNames}>";
         }
--- a/Logging/.CSProject/Logger/Logger.cs
+++ b/Logging/.CSProject/Logger/Logger.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Linq;
 using System.Runtime.CompilerServices;
 
 namespace Anvil.CSharp.Logging
@@ -14,6 +15,33 @@ namespace Anvil.CSharp.Logging
     /// </summary>
     public readonly struct Logger : ILogger
     {
+        // NOTE: This is a duplicate of Anvil.CSharp.Reflection.TypeExtension.GetReadableName
+        // because the logging namespace is isolated from the rest of Anvil, and thus cannot access that extension
+        private static readonly Type s_NullableType = typeof(Nullable<>);
+        private static string GetReadableName(Type type)
+        {
+            if (!type.IsGenericType)
+            {
+                return type.Name;
+            }
+
+            if (type.GetGenericTypeDefinition() == s_NullableType)
+            {
+                return $"{GetReadableName(type.GenericTypeArguments[0])}?";
+            }
+
+            // Remove the generic type count indicator (`n) from the type name
+            // Avoid having to calculate the number of digits in the generic type count by assuming it's under 100
+            int removeCount = 1 + (type.GenericTypeArguments.Length < 10 ? 1 : 2);
+            string name = type.Name[..^removeCount];
+
+            string genericTypeNames = type.GenericTypeArguments
+                .Select(GetReadableName)
+                .Aggregate((a, b) => $"{a}, {b}");
+
+            return $"{name}<{genericTypeNames}>";
+        }
+
         /// <summary>
         /// The name of the type this <see cref="Logger"/> represents.
         /// </summary>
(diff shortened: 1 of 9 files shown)


jkeon (dismissed):
Yay!

tjvezina:
Pausing so I can take the opportunity to add unit tests when https://github.com/decline-cookies/anvil-csharp-core/pull/117 is merged, and maybe a GitHub Action to run them!

mbaker3:
I made the edit but just a heads up that this part of the PR template is intended to have you reference resolved issues (if any)
image

Logging/.CSProject/Logger/Logger.cs
```diff
@@ -14,6 +15,33 @@ namespace Anvil.CSharp.Logging
...
+            }
+
+            // Remove the generic type count indicator (`n) from the type name
+            // Avoid having to calculate the number of digits in the generic type count by assuming it's under 100
+            int removeCount = 1 + (type.GenericTypeArguments.Length < 10 ? 1 : 2);
+            string name = type.Name[..^removeCount];
+
+            string genericTypeNames = type.GenericTypeArguments
+                .Select(GetReadableName)
+                .Aggregate((a, b) => $"{a}, {b}");
```
mbaker3, line 40: Lots of string allocations concatenating like this.

Fewer allocs:
```suggestion
            IEnumerable<string> readableArgs = type.GenericTypeArguments
                .Select(GetReadableName);
             string genericTypeNames = string.Join(readableArgs, ", ");
```

If we wanted to take it further we'd pass a `StringBuilder` through the recursion but that's probably excessive until we prove that logger is hammering the GC.

Reflection/TypeExtension.cs
```diff
@@ -32,5 +35,36 @@ public static bool isStatic(this Type type)
...
         }
+
+        /// <summary>
+        /// Gets a human-readable type name, similar to how the type appears in code. Primarily handles generic types.
+        /// For example, instead of "List`1" or "System.Collections.Generic.List`1[System.Int32]", this helper will
+        /// return the name "List<Int32>"
+        /// </summary>
+        /// <param name="type">The type to get a readable name for.</param>
+        /// <returns>The readable type name.</returns>
+        public static string GetReadableName(this Type type)
```
mbaker3, line 46:
```suggestion
        <remarks>This logic is duplicated in <see cref="Logger" /> until the dll can be merged back into the main anvil library.</remarks>
        public static string GetReadableName(this Type type)
```

mbaker3 (changes requested):

tjvezina:
> I made the edit but just a heads up that this part of the PR template is intended to have you reference resolved issues (if any)

Ohh that makes sense 😆 I took it to literally mean problems, not GitHub Issues, good to know!

mbaker3 (dismissed):

Tests/anvil-csharp-tests.asmdef
```diff
@@ -1,7 +1,9 @@
 {
     "name": "anvil-csharp-tests",
     "rootNamespace": "Anvil.CSharp.Tests",
-    "references": [],
+    "references": [
+        "anvil-csharp-core"
+    ],
```
tjvezina, line 6: @jkeon You were right - a Unity assembly definition will auto-reference DLL's, but not other assembly definitions. I inadvertently tested this unit test project by writing a Logging test... which I was only able to access because it's already compiled into a DLL! This reference makes the rest of Anvil accessible (in Unity).

mbaker3 (approved):

Tests/anvil-csharp-tests.asmdef
jkeon: Ok good to know!

PR #123, opened 2022-10-14
Handle log caller file paths in a platform-independent way

tjvezina:
- Replace `Path.GetFilePathWithoutExtension` with explicit string splitting, to parse file paths independent of OS
- Bundled log source details in a `CallerInfo` struct passed to all handlers
    - Caller file name is extracted from the full file path in `Log.DispatchLog` rather than requiring each handler to do so
- Changed `Debug.Assert(!IsHandlingLog)` to a thrown exception, to ensure checking in release builds
- Changed unknown log context string from `unknown` to `<unknown>`
- Changed `Logger.DerivedTypeName` to `Logger.OwnerTypeName`

On Mac, if the logging DLL was most recently compiled on Windows, logs from within the DLL give a Windows file path, which `Path.GetFileNameWithoutExtension` fails to parse, returning the full path:
```
(C:\Users\Tyler\Projects\Blacksmith\Assets\Anvil\AnvilCSharp\Logging\.CSProject\Log:81|InitLogListeners) Default Log listener UnityLogListener initialized.
```

Caller file names are correctly extracted, independent of OS or path separator:
```
(Log:81|InitLogListeners) Default Log listener UnityLogListener initialized.
```

- None

 - None

 - [X] Yes
     - Any `ILogHandler` defined outside of Anvil will need to be updated to use the new `CallerInfo` argument
 - [ ] No


tjvezina:
New plan after an offline discussion: pass a `readonly struct` to log handlers instead of multiple args, and include both the full path and file name.

tjvezina:
Another draft of suggested changes (without documentation until finalized) - I took issue with the name of `Logger.DerivedTypeName`/`callerDerivedTypeName` as there's nothing to say the caller of a log must also be the type originally passed to the `Logger`. I'm open to suggestions, but for now changed it to `OwnerTypeName` and `CallerInfo.LoggerName` to represent that "this is the name/type given to the Logger instance (but not necessarily the source of the log message)".

If the type given to the `Logger` is meant to always match the log source, I would question its usefulness? It's not used in any handlers within Anvil, is it being used elsewhere? Could it be replaced with the `[CallerMethodName]`'s declaring type instead, or just removed?

Initially, I thought `GetStaticLogger(typeof(MyType))` would return the same `Logger` instance every time, but it's always a new one, and the type passed in isn't used by default. Could `Logger` be wholly removed in favor of the `[Caller<Info>]` attributes given when a log is sent? Then you could simply `Log.Debug(...)` anywhere, and you'd still get the same log with the same context, no?

jkeon (commented):
This looks good but I'll let @mbaker3 make the call since you two have been deep in logger territory.

> If the type given to the Logger is meant to always match the log source, I would question its usefulness? It's not used in any handlers within Anvil, is it being used elsewhere? Could it be replaced with the [CallerMethodName]'s declaring type instead, or just removed?

I'm not sure if this is what you mean or not but one thing I've noticed is that the Filename isn't super useful in the context of inheritance. 
Screen Shot 2022-10-14 at 9 51 02 AM

This tells me where the log was emitted which is great but I really want to know who spit that log out. 

`(TimerTaskSystem|OnCreate:60) Test`
`(MoveToTaskSystem|OnCreate:60) Test`
etc

Of course, the file line number no longer makes sense in that context since it's the type and not the file. 

Maybe the type is included unless it matches the filename? 

`(TimerTaskSystem - AbstractTaskSystem|OnCreate:60) Test log in base class`
`(TimerTaskSystem|ConfigureJobs:18) Test log in same file`

I open it up for discussion.

While we're discussing, I also noticed this:
Screen Shot 2022-10-14 at 10 01 12 AM

Is this API intended? Seems like the public facing one should only have message? And the other parts are hidden internally?

tjvezina:
> This tells me where the log was emitted which is great but I really want to know who spit that log out.

That makes sense - if the log was emitted from a base class function, that base class's file path is sent to the log. Maybe that was the original reason for supplying an explicit type/instance to `Logger`, except again that type name is not used by default in any log handlers, the file path is instead. I would personally rather see that derived/owner type name than the file or line number. It could be both as you suggested, but the log context starts to get awfully long then. Maybe the file/line could be appended on a new line instead?

> Is this API intended?

Yes, because each of those args has a special attribute (ex. `[CallerMethodName] string callerName`) meaning they are auto-populated if not explicitly set. It's weird that they're visible, but the result is that when you simply call `Logger.Debug(message)` the function is magically given the caller's file path, method name, and line number too.

tjvezina:
At this point, there has been enough confusion between file names and type names in log contexts that I'd prefer to include the file extension and be clear. Maybe like this:
```
(AbstractTaskSystem.cs:60|TimerTaskSystem.OnCreate) Test
```
...omitting the `TimerTaskSystem.` part if it's the same as the file name.

jkeon:
> That makes sense - if the log was emitted from a base class function, that base class's file path is sent to the log. Maybe that was the original reason for supplying an explicit type/instance to Logger, except again that type name is not used by default in any log handlers, the file path is instead. I would personally rather see that derived/owner type name than the file or line number. It could be both as you suggested, but the log context starts to get awfully long then. Maybe the file/line could be appended on a new line instead?

Yeah I'm not sure the best solution here, I'd be fine with a new line to provide that additional context. I know @mbaker3 was just saying he uses the console on a one line view so he wouldn't see it unless he clicked in... but maybe that's fine if the filename and line are what are hidden below the fold? 

See the type context and method plus your log for first line usage.
See the filename and line number on second line to drill in.

> Yes, because each of those args has a [special attribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callermembernameattribute?view=net-6.0) (ex. [CallerMemberName] string callerName = "") meaning they are auto-populated if not explicitly set. It's weird that they're visible, but the result is that when you simply call Logger.Debug(message) the function is magically given the caller's file path, method name, and line number too.

Yeah it's fine because I know how to use it, but every so often when I see intellisense come up (and especially the first time) I felt like I didn't have the right API and I felt like I should manually populate those fields.

jkeon:
Fair point on the filename extension!

Logging/.CSProject/Log.cs
```diff
@@ -163,21 +163,24 @@ private static bool ShouldIgnoreAssembly(Assembly assembly)
...
+            if (IsHandlingLog)
+            {
+                throw new Exception("A log is already being handled");
+            }
+
+            CallerInfo callerInfo = new CallerInfo(
+                loggerName,
+                callerName ?? UNKNOWN_CONTEXT,
+                callerPath ?? UNKNOWN_CONTEXT,
+                (!string.IsNullOrEmpty(callerPath) ? callerPath.Split('/', '\\').Last().Split('.').First() : UNKNOWN_CONTEXT),
```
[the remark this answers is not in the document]
mbaker3: yea.... Should probable do a couple of `LastIndexOf` searches and a single substring.

mbaker3, line 177: Should we just do this check once and deal with callerPath and callerFile at the same time?

Logging/.CSProject/Log.cs
```diff
@@ -163,21 +163,24 @@ private static bool ShouldIgnoreAssembly(Assembly assembly)
...
             IsHandlingLog = true;
             foreach (ILogHandler handler in s_AdditionalHandlerList)
             {
-                handler.HandleLog(level,
-                message,
-                callerDerivedTypeName,
-                callerPath,
-                callerName,
-                callerLine);
+                handler.HandleLog(level, message, callerInfo);
```
mbaker3, line 183:
```suggestion
                handler.HandleLog(level, message, in callerInfo);
```

Not necessary but stylistically preferred

Logging/.CSProject/Logger/Logger.cs
```diff
@@ -8,9 +8,10 @@ namespace Anvil.CSharp.Logging
     /// A context specific instance that provides a mechanism to emit logs through <see cref="Log"/>.
     /// Automatically provides contextual information to <see cref="Log"/> about caller context including:
     ///  - Optional, per instance, message prefix
-    ///  - Caller type name
+    ///  - Logger's owner's type name
+    ///  - Caller method name
     ///  - Caller file path
-    ///  - Caller name
+    ///  - Caller file name
```
mbaker3, line 14: File name isn't provided by Logger. Just the path.

Caller type name is though

mbaker3 (commented):
> Changed Debug.Assert(!IsHandlingLog) to a thrown exception, to ensure checking in release builds

This is OK, since we now compile the DLLs in release mode but the intent was that the only time this will trigger is when developing a new Log Handler or Listener. It's really unlikely it will ever happen in release build.

That said, it's a cheap check so it's not the end of the world to have it in release builds.

> Another draft of suggested changes (without documentation until finalized) - I took issue with the name of Logger.DerivedTypeName/callerDerivedTypeName as there's nothing to say the caller of a log must also be the type originally passed to the Logger. I'm open to suggestions, but for now changed it to OwnerTypeName and CallerInfo.LoggerName to represent that "this is the name/type given to the Logger instance (but not necessarily the source of the log message)".

Agreed that there isn't anything that enforces calling type == derivedTypeName but that is the intended use. When you call `Log.GetLogger(instance)` I'm not sure what else people would expect to receive other than a logger tailored to the context of the instance they provided.

I'm not sure how you'd enforce correct usage.

> If the type given to the Logger is meant to always match the log source, I would question its usefulness? It's not used in any handlers within Anvil, is it being used elsewhere?

Generally, yes the Logger is always meant to match the top level instance type. The derived type is not used by default in any handlers at the moment but it is important to have available as an option when configuring the log formatting of a handler. Today only `FileLogHandler` allows this kind of customization but the intent is to allow it across all handlers.

In fact, the derived type name is the whole reason `Logger` exists. To provide top level type context to logs on projects that wanted the ability to include it in log messages! `messagePrefix` was just a tag along feature!

We do use non-default formatting and the derived type name in at least one project.


> Could it be replaced with the [CallerMethodName]'s declaring type instead, or just removed?

How would this work? That attribute just populates a string on the parameter.


### Re Changing Default Log Format
I'm not opposed to changing the default log formats to lean on the derived type rather than the file name. There are pros and cons to using type vs file name. If you don't use file name then line number isn't useful. However, without filename/line number it can be difficult to hunt down the exact call site when there are multiple logs in a method and the message is poorly designed.
Ex: `Logger.Debug(myBool)`

And I hear you already, "Just design your log messages better". The reality is that logs aren't given much thought and you typically don't realize they're poor until you're reading through a log trying to figure out what happened. It's too late at that point.

The below the fold vs above the fold discussion is only relevant to the `UnityLogHandler`. Because of the way the Unity console handles multi-line logs I think a format like the following would work well:
```
(TimerTaskSystem.OnCreate) Test
(at AbstractTaskSystem.cs:60)
```
This matches the way the stack is written by Unity in the lines after. Really the file and line are a bit redundant since they're going to be the first blue text in the stack and the destination you arrive at when double clicking the message.

All other log handlers should emit one line logs by default. In those other handlers
 - You typically have much more screen real estate
 - Visually parsing multi-line logs is awful aside from exceptional cases where you're dumping JSON

For those handlers I like @tjvezina's suggestion but I wouldn't remove the type if it matches the file name. Visual and semantic consistency in log formatting is more important for both readability and parse-ability. 
> `(AbstractTaskSystem.cs:60|TimerTaskSystem.OnCreate) Test`

Maybe reverse the content on either side of the `|` so type/method comes before file/number.

> > Yes, because each of those args has a [special attribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callermembernameattribute?view=net-6.0) (ex. [CallerMemberName] string callerName = "") meaning they are auto-populated if not explicitly set. It's weird that they're visible, but the result is that when you simply call Logger.Debug(message) the function is magically given the caller's file path, method name, and line number too.
> 
> Yeah it's fine because I know how to use it, but every so often when I see intellisense come up (and especially the first time) I felt like I didn't have the right API and I felt like I should manually populate those fields.

Unfortunately there's no real way around this. There weren't complaints when this was originally introduced :) #62

Logging/.CSProject/CallerInfo.cs
```diff
@@ -0,0 +1,26 @@
...
+    public readonly struct CallerInfo
+    {
+        public readonly string LoggerName;
+        public readonly string MethodName;
+        public readonly string FullPath;
+        public readonly string FileName;
+        public readonly int LineNumber;
+
+        public CallerInfo(
+            string loggerName,
```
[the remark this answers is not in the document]
tjvezina: I feel like I followed in your footsteps and ended up back in a spot where `callerDerivedTypeName` makes the most sense... from an outside perspective it feels inaccurate, but we're making the assumption that your custom logger (i.e. `Log.GetLogger(this)`) will only be used within the instance that created it...

I imagined creating a logger, then accessing it from related files/components to group the logs together, resulting in a `callerDerivedTypeName = "Player"` but a `callerPath = "HealthBar.cs"` - but maybe that's less likely than I expect.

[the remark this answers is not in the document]
mbaker3: It's a journey :)

Yea I don't think sharing logger instances should be encouraged. `AbstractAnvilBase.Logger` probably represents 90% of usage anyway.

For grouping I think `messagePrefix` would be helpful. Or introducing some sort of tagging feature.

Logging/.CSProject/Log.cs
```diff
@@ -142,18 +142,18 @@ private static bool ShouldIgnoreAssembly(Assembly assembly)
         /// <summary>
         /// Creates a <see cref="Logger"/> instance for a given instance.
         /// </summary>
-        /// <param name="instance">The instance to create the <see cref="Logger"/> instance for.</param>
+        /// <param name="owner">The object to create a <see cref="Logger"/> instance for.</param>
```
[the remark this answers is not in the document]
tjvezina: It just seemed more descriptive to refer to the given object as the Logger's "owner" rather than simply an "instance" (I mean, everything that isn't a primitive is an object instance, right?) I can change it back to `instance` though, just trying some things to see what fit best.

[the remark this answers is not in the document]
mbaker3: oh `owner` is fine as a param. That or `context`. 
I was more referring to the use of instance in the comment.

If instance is everything non-primitive then object is everything including primitive!

tjvezina:
> > Changed Debug.Assert(!IsHandlingLog) to a thrown exception, to ensure checking in release builds
>
> [...] It's really unlikely it will ever happen in release build.

I figured Anvil consumers could define their own `ILogHandler`, and try to do something tricky (or accidental during development), inadvertently sending a log while already handling a log, so you'd still want an exception thrown.

> > Could it be replaced with the [CallerMethodName]'s declaring type instead, or just removed?
> 
> How would this work? That attribute just populates a string on the parameter.

I realized after I was thinking about the StackFrame business in `UnityLogListener`, which doesn't apply to most logs - it would have to be the manually supplied type in the `Logger` constructor.

I didn't realize `AbstractAnvilBase` included a `Logger` property, that does make accessing easier with proper Anvil use. I was getting sick of `Log.GetLogger(this).Debug(...)` every time I log something in my demo project.

> The below the fold vs above the fold discussion is only relevant to the UnityLogHandler. 

Makes sense - I use single-line view in the Unity console too, but wouldn't mind knowing the file/line are on the next line if I want to select that log and see them. I'll try that out!

> Maybe reverse the content on either side of the | so type/method comes before file/number.

👍

Logging/.CSProject/Log.cs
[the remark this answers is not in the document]
tjvezina: As per discussions below, if we're keeping the file extension I can drop half of this and just take the substring after the final path separator, easy.

Logging/.CSProject/Logger/Logger.cs
```diff
@@ -52,87 +52,87 @@ private static string GetReadableName(Type type)
...
             LogLevel.Debug,
             string.Concat(MessagePrefix, message),
-            DerivedTypeName,
-            callerPath,
-            callerName,
-            callerLine);
+            OwnerTypeName,
+            callerMethodName,
+            callerFilePath,
+            callerLineNumber);
```
tjvezina, line 89: I hesitated to reorganize these, but the consistency is nice to have, and I assume there aren't a lot of cases outside Anvil that were supplying custom values for these args, or few enough they can be easily updated.

tjvezina:
Updated some variable names and the default log context format. Some examples (from a Unity project):
```
(Log.InitLogListeners|Log.cs:81) Default Log listener UnityLogListener initialized.
(ID_StaticStateResetHelper.Init|ID_StaticStateResetHelper.cs:17) Resetting ID state.
(HybridRendererSystem.OnCreate|HybridV2RenderSystem.cs:1729) No SRP present, [...]
(Demo.TestDiagnostics|Demo.cs:102) write message
(Demo.TestDiagnostics|Demo.cs:124) UnityEngine.Assertions.AssertionException: An assertion failed
(UnityLogListener..ctor|<unknown>) System.Exception: There can only be one instance [...]
```
I'm still going to try moving the file name/line number to a new line in UnityLogHandler (but that would be part of https://github.com/decline-cookies/anvil-unity-core/pull/89) and may take a stab at moving the customizable log context format up from `FileLogHandler` to all handlers.

Logging/.CSProject/LogHandler/AbstractLogHandler.cs
```diff
@@ -0,0 +1,102 @@
...
+        protected AbstractLogHandler()
+        {
+            LogFormat = DefaultLogFormat;
+        }
+
+        internal void HandleLog(LogLevel level, string message, in CallerInfo callerInfo)
+        {
+            HandleFormattedLog(level, string.Format(
+                m_LogFormat,
+                message,
```
tjvezina, line 84: I marked this (and `CallerInfo`) `internal` as they are entirely contained within the logging assembly. If keeping this public is preferred, either so `myLogHandler.HandleLog(...)` can be called directly or so `callerInfo` can be forwarded to `HandleFormattedLog` too, let me know and I can change it, but it felt cleaner to keep the log formatting details wrapped up within Anvil, and simply pass a `LogLevel` and formatted string to handlers.

Logging/.CSProject/LogHandler/AbstractLogHandler.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// <summary> Log format part - the caller's source file path </summary>
+        public const string LOG_PART_CALLER_PATH = "{3}";
+        /// <summary> Log format part - the caller's source file name </summary>
+        public const string LOG_PART_CALLER_FILE = "{4}";
+        /// <summary> Log format part - the caller's source file line number </summary>
+        public const string LOG_PART_CALLER_LINE = "{5}";
+        /// <summary> Log format part - the timestamp at which the log was emitted. See <see cref="TimestampFormat"/> to customize format. </summary>
+        public const string LOG_PART_TIMESTAMP = "{6}";
+        /// <summary> Log format part - the log level. </summary>
+        public const string LOG_PART_LOG_LEVEL = "{7}";
```
jkeon, line 25: Interesting  I didn't know you could opt into these with string.Format. I assumed it was more "dumb" and expected params in order {0}, {1}, {etc} and that the number of actual values you passed in had to match.

jkeon (approved):
Looks good!

Logging/.CSProject/LogHandler/AbstractLogHandler.cs
jkeon: I think that makes sense to me

Logging/.CSProject/LogHandler/AbstractLogHandler.cs
tjvezina: It does still work that way, the `string.Format(...)` in `HandleLog` is passed each of the strings that correspond to these constants every time, so their position doesn't change (ex. the original message is always passed first).

jkeon: Yeah sorry I meant I didn't think you could have "holes" in your formatting. 

If you pass a string with {0}, {3}, {5} I didn't think that would work.

tjvezina: Oh gotcha, yeah they don't have to all be present, or even in order or only included once.

Logging/.CSProject/Logger/Logger.cs
mbaker3: yea, all good. Just need to be super careful for `UnityLogListener` and the like that manually supply these values

Logging/.CSProject/LogHandler/AbstractLogHandler.cs
```diff
@@ -0,0 +1,102 @@
+﻿using System;
+
+namespace Anvil.CSharp.Logging
+{
+    /// <summary>
+    /// A type capable of handling logs received by <see cref="Log"/>.
+    /// </summary>
+    public abstract class AbstractLogHandler
```
mbaker3, line 8: I think I'm ok with this but a couple thoughts came to mind.

### Forcing base type
Now that we're requiring that handlers derive from a specific type. Are we concerned about potential platforms/contexts that would require a different base type? Say a specific type in order to submit logs to their system

### Redundant options
With a common set of formatting options it's a bit weird that you can select redundant options for a handler.
For example, log level and timestamp are baked into Unity's logging API so they aren't really valid options for that handler.

---

If we did want to go back to `ILogHandler` but consolidate the log formatting logic we could create a `LogFormat` or `LogFormatter` type that most handlers would hold as an instance 🤷

Like I said, I'm fine with the change but I just wanted to express the uneasiness I was feeling in case you guys hadn't thought about these specific angles.

Logging/LogHandler/FileLogHandler.cs
```diff
@@ -151,45 +114,28 @@ private void CreateWriter()
...
-            string callerDerivedTypeName,
-            string callerPath,
-            string callerName,
-            int callerLine)
+        protected override void HandleFormattedLog(LogLevel level, string formattedLog)
         {
             if ((int)level < (int)MinimumLevel)
             {
                 return;
             }
```
mbaker3, line 127: If we do keep `AbstractLogHandler` we should drop this down into the type. No sense in doing all the work to format a log only to bail out. This is a useful concept for all log handlers.

Logging/.CSProject/LogHandler/AbstractLogHandler.cs
mbaker3: In its current form this is fine but depends on the outcome of this comment https://github.com/decline-cookies/anvil-csharp-core/pull/123#discussion_r997490364

Logging/LogHandler/FileLogHandler.cs
tjvezina: I was on the fence about that, since setting a minimum log level seems specific to file writing - you definitely want the option to see only errors in a big log file, but you'd never prevent debug logs/warnings from being sent to any console/editor/IDE would you?

Logging/.CSProject/LogHandler/AbstractLogHandler.cs
tjvezina: > ### Forcing base type
 
I did have to change `FileLogHandler` from an `AbstractAnvilBase` to simply `IDisposable`, but I don't think that's a problem - it doesn't need the extra `DisposeSelf` or `Logger` functionality anyway.

We currently have several log handlers but no instances of this problem, and at worst if it comes up later we either rewrite this again, or you use composition instead of inheritance for that special case.

> ### Redundant options

I don't think flexibility is a bad thing. Maybe you don't like how Unity formats their timestamp, or its position in the log, so you want to customize it. Maybe you use a different IDE/C# tool that already includes caller info in the console, so you remove those in your custom handler.

The alternative would be to tailor the context options to each handler, which seems to defeat the purpose of automatically handling log formatting for handlers out of the box.

> `LogFormatter`

If the base is just an interface, how does it provide and use a `LogFormatter` instance? I guess recent C# versions allow default implementations but I'm not sure if you could achieve the same thing. 🤔 If you've got an idea for how to make it work I'd be happy to implement.

Logging/LogHandler/FileLogHandler.cs
mbaker3: For filtering purposes potentially. Unity's console has that filtering functionality baked in but it filters from all logs.

I know we had a custom log handler in our console application for a bit that only showed a subset of messages to the user. I ended up simplifying and just monitoring the log file directly then parsing out the log level from the message but that was more out of laziness than implementing the ideal approach. A custom handler was probably the best way to go if I had more time.

I could also see restricting log levels in release builds vs debug builds for performance reasons or to prevent logging sensitive data.

Logging/.CSProject/LogHandler/AbstractLogHandler.cs
mbaker3: ok, all good. Was just my 2 cents. No need for changes

> I did have to change FileLogHandler from an AbstractAnvilBase to simply IDisposable, but I don't think that's a problem - it doesn't need the extra DisposeSelf or Logger functionality anyway.

Saw that. I initially thought of asking you to implement `IAnvilDisposable` but it's really not necessary for this use case. The disposal is very straight forward and controlled.

> > `LogFormatter`
> 
> If the base is just an interface, how does it provide and use a `LogFormatter` instance? I guess recent C# versions allow default implementations but I'm not sure if you could achieve the same thing. 🤔 If you've got an idea for how to make it work I'd be happy to implement.

I was thinking the concrete implementation (`FileLogHandler`, `ConsoleLogHandler`, etc..) would just create an instance of `LogFormatter` rather than pulling it from a default implementation. Maybe the developer would even pass in a pre-configured LogFormatter.

mbaker3 (approved):

PR #124, opened 2022-10-15
Added unit tests for `Array2DExtension`

tjvezina:
- #105

 - None

 - [ ] Yes
 - [X] No

### DIFF
--- a/Tests/Data/Collections/Util/Array2DExtensionTests.cs
+++ b/Tests/Data/Collections/Util/Array2DExtensionTests.cs
@@ -1,0 +1,94 @@
+using NUnit.Framework;
+using System;
+using Anvil.CSharp.Data;
+
+namespace Anvil.CSharp.Tests
+{
+    public static class Array2DExtensionTests
+    {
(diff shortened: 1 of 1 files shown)


Tests/Data/Collections/Util/Array2DExtensionTests.cs
```diff
@@ -0,0 +1,94 @@
+using NUnit.Framework;
+using System;
+using Anvil.CSharp.Data;
+
+namespace Anvil.CSharp.Tests
+{
+    public static class Array2DExtensionTests
+    {
+        [Test, Order(0)]
```
tjvezina, line 9: This `[Order]` attribute just ensures the `PopulateTest` runs before the rest, as most depend on that function. Tests without this attribute always run after those that do.

tjvezina:
Unrelated: Is the "What issues does this resolve?" header in the PR template necessary, given GitHub already has an interface for associating issues with PR's?

jkeon (approved):

Issue #102, opened 2022-08-16
Setup CodeQL


mbaker3:
**Related:**

Setup CodeQL analysis using GitHub Actions:
https://github.com/decline-cookies/anvil-csharp-core/security/code-scanning

See test implementation at
https://github.com/mbaker3/anvil-csharp-core/blob/main/.github/workflows/codeql-analysis.yml

Make sure Logging and core projects are analyzed using both their dot net core and msbuild versions.

mbaker3:
Partially completed by #125 but requires additional work outlined in this comment
https://github.com/decline-cookies/anvil-csharp-core/pull/125#pullrequestreview-1144746780

PR #125, opened 2022-10-15
Add CodeQL GitHub Actions workflow

tjvezina:
- #102

 - None

 - [ ] Yes
 - [X] No

---

Just wanted to open the PR and see if this workflow would succeed!

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -65,13 +65,13 @@ jobs:
     #   modify them (or add more) to build your code if your project, please refer to the EXAMPLE below for guidance.
 
     - name: Build Logging
-    - run: |
+      working-directory: ./Logging/.CSProject
+      run: |
         echo "Building logging project"
         dotnet build anvil-csharp-logging.sln
-    - working-directory: ./Logging/.CSProject
         
     - name: Build Core
-    - run: |
+      run: |
         echo "Building core project"
         dotnet build anvil-csharp-core.csproj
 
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,79 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "main" ]
+  schedule:
+    - cron: '37 9 * * 6'
```
tjvezina, line 21: Saturdays at 9:37 AM UTC is awfully specific

jkeon (approved):

.github/workflows/codeql-analysis.yml
mbaker3: Yea, when you have GitHub create the workflow it seems to randomly select a random weekly time. Probably to prevent everyone from hammering their CodeQL runners at the same time every week.

I just created another workflow now and it gave me `- cron: '28 13 * * 2'`

tjvezina: Ah that makes sense! Well I just wanted to see if the workflow would succeed (it did), feel free to take over this PR or start a new one since they're your changes.

mbaker3 (approved):
Approved but I don't think this resolves #102. Admittedly that task isn't well written but the intent was to have code quality and security analyzed. I just added 38c1055 which enables the quality scanning but I think we need to filter which rules we want to surface to complete the task.

For example, I don't think we want to warn on this rule: https://github.com/decline-cookies/anvil-csharp-core/security/code-scanning/18

---

Also, the `set-output` warning emitted during the action. Looks like the warning will be fixed by GitHub in the next release https://github.com/github/codeql-action/issues/1305

Issue #126, opened 2022-10-20
Logging - Make thread safe


mbaker3:
The logging system is not currently thread safe. This means that all calls into `Logger` and `Log` must happen on the same thread.

`Log` should probably be the sync point for calls since most `AbstractLogHandler` destinations are unlikely to be thread safe and it'll be unreasonable/inconsistent to have them all implement their own marshalling logic.

There are currently `ILogListener` implementations that marshal logs onto a single thread (Ex: [UnityLogListener](https://github.com/decline-cookies/anvil-unity-core/blob/main/Scripts/Runtime/Logging/.CSProject/UnityLogListener.cs)) but this is an incomplete solution that only catches calls that aren't initiated through a `Logger` instance. 

At the moment, the developer is burdened with marshalling calls made directly through `Logger`. This really hampers the convenience of `Logger` in a multi-threaded context.

## Steps
- [ ] Implement thread safety in `Log`
   - Think through and discuss pros/cons of locking/SlimSemaphore vs building a concurrent queue and pumping messages
- [ ] Remove thread marshalling from existing `ILogListener` implementations

PR #127, opened 2022-10-20
Hotfix logging

mbaker3:
`OneTimeLogger` - Fix pass through call site parameter order that was missed in #123 

## Tag Along
Added references to #126 and the lack of thread safety when using logger. We just encountered this in an app.

Any logging through `OneTimeLogger` has the method name and file path swapped in log messages.

Log messages through `OneTimeLogger` are now formatted consistently with all other loggers.

None

 - None

 - [ ] Yes 
 - [x] No


tjvezina (approved):
I didn't even realize (or forgot) that `OneTimeLogger` is a thing!

PR #130, opened 2022-10-21
TypeExtension.GetReadableName Improvements

mbaker3:
Improved `TypeExtension.GetReadableName` to:
 - Include outer types in the names of inner types.
 - Correctly handle types with a generic outer type - `MyType<int>.MyInnerType`

## TODO
Once this is approved there is some work to do copying the changes.
 - [x] ~Copy `GetReadableName()` logic into the Logging DLL~

## Tag Alongs
 - `MathUtil.GetDigitCount` - Returns the number of digits in an int without needing to convert to string.
 - Add unit tests for `MathUtil`
    - For the prime number test I added some reasonable testing that should hit all code paths but checking that the first `int.MaxValue` prime numbers are correct is just making unreasonable work for the unit tests!

`GetReadableName`
 -  Does not include the names of outer types on a given type.
 - Would generate an incorrect name when an inner type had a generic outer type
    - Ex: `MyType<int>.MyInnerType` would resolve to `MyInnerTy<int>`

 - Outer types are included in the generated name
 - All combinations of generic and inner types are handled correctly
    - `MyType<int>.MyInnerType`
    - `MyType<int>.MyInnerType<float>`
    - `MyType.MyInnerType<float>`

None

 - None

 - [x] Yes
 - [ ] No

Migration instructions:
 - Any use of `TypeExtension.GetReadableName()` must now add `using Anvil.CSharp.Logging`
 - Any use of `MathUtil.GetDigitCount()` must now add `using Anvil.CSharp.Logging` and convert the use to an extension method.
    - Ex: `MathUtil.GetDigitCount(13)` is now `13.GetDigitCount()`

Breaking changes in other Anvil repos that depend on this one:
 - https://github.com/decline-cookies/anvil-unity-dots/pull/96

### DIFF
--- a/Mathematics/MathUtil.cs
+++ b/Mathematics/MathUtil.cs
@@ -43,44 +43,5 @@ namespace Anvil.CSharp.Mathematics
 
             return (--a);
         }
-
-        /// <summary>
-        /// Returns the number of digits in an integer.
-        /// Optionally, includes the negative sign in the count.
-        /// </summary>
-        /// <param name="value">The value to evaluate</param>
-        /// <param name="countSign">(default:false) If true, will increase the count by one for all numbers less than 0</param>
-        /// <returns>The number of digits in an integer</returns>
-        /// <remarks>Taken from: https://stackoverflow.com/a/51099524/640196</remarks>
-        public static int GetDigitCount(int value, bool countSign = false)
-        {
-            if (value >= 0)
-            {
-                if (value < 10) return 1;
-                if (value < 100) return 2;
-                if (value < 1_000) return 3;
-                if (value < 10_000) return 4;
-                if (value < 100_000) return 5;
-                if (value < 1_000_000) return 6;
-                if (value < 10_000_000) return 7;
-                if (value < 100_000_000) return 8;
-                if (value < 1_000_000_000) return 9;
-                return 10;
-            }
-            else
-            {
-                int signCount = countSign ? 1 : 0;
-                if (value > -10) return signCount + 1;
-                if (value > -100) return signCount + 2;
-                if (value > -1_000) return signCount + 3;
-                if (value > -10_000) return signCount + 4;
-                if (value > -100_000) return signCount + 5;
-                if (value > -1_000_000) return signCount + 6;
-                if (value > -10_000_000) return signCount + 7;
-                if (value > -100_000_000) return signCount + 8;
-                if (value > -1_000_000_000) return signCount + 9;
-                return signCount + 10;
-            }
-        }
     }
 }
--- a/Mathematics/MathUtil.cs
+++ b/Mathematics/MathUtil.cs
@@ -58,28 +58,28 @@ namespace Anvil.CSharp.Mathematics
             {
                 if (value < 10) return 1;
                 if (value < 100) return 2;
-                if (value < 1000) return 3;
-                if (value < 10000) return 4;
-                if (value < 100000) return 5;
-                if (value < 1000000) return 6;
-                if (value < 10000000) return 7;
-                if (value < 100000000) return 8;
-                if (value < 1000000000) return 9;
+                if (value < 1_000) return 3;
+                if (value < 10_000) return 4;
+                if (value < 100_000) return 5;
+                if (value < 1_000_000) return 6;
+                if (value < 10_000_000) return 7;
+                if (value < 100_000_000) return 8;
+                if (value < 1_000_000_000) return 9;
                 return 10;
             }
             else
             {
                 int signCount = countSign ? 1 : 0;
-                if (value > -10) return signCount + 2;
-                if (value > -100) return signCount + 3;
-                if (value > -1000) return signCount + 4;
-                if (value > -10000) return signCount + 5;
-                if (value > -100000) return signCount + 6;
-                if (value > -1000000) return signCount + 7;
-                if (value > -10000000) return signCount + 8;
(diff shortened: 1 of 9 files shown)


Mathematics/MathUtil.cs
```diff
@@ -43,5 +43,44 @@ public static long FindPrimeNumber(int nthPrimeNumberToFind)
...
+                if (value < 100) return 2;
+                if (value < 1_000) return 3;
+                if (value < 10_000) return 4;
+                if (value < 100_000) return 5;
+                if (value < 1_000_000) return 6;
+                if (value < 10_000_000) return 7;
+                if (value < 100_000_000) return 8;
+                if (value < 1_000_000_000) return 9;
+                return 10;
+            }
```
tjvezina, line 69: Could you just use `Math.Log10`?
```cs
return Math.Floor(Math.Log10(Math.Abs(value))) + 1
```
or is this far more performant or something?

Reflection/TypeExtension.cs
```diff
@@ -48,24 +51,98 @@ public static bool isStatic(this Type type)
...
+        /// <param name="type">The type to get a readable name for.</param>
+        /// <param name="genericArgs">
+        /// The generic arguments of the type ordered from outer to inner generic arguments.
+        /// Ex: MyType<genericArgs[0]>.MyInnerType<genericArgs[1]>
+        /// (The default order returned by <see cref="Type.GenericTypeArguments"/>
+        /// </param>
+        /// <param name="genericArgsUpperBound">
+        /// The upper bounds (exclusive) of the <see cref="genericArgs"/> array to evaluate
+        /// </param>
+        /// <returns></returns>
```
tjvezina, line 82: Blank return value description

Reflection/TypeExtension.cs
```diff
@@ -48,24 +51,98 @@ public static bool isStatic(this Type type)
...
             }
 
             if (type.GetGenericTypeDefinition() == s_NullableType)
             {
-                return $"{type.GenericTypeArguments[0].GetReadableName()}?";
+                return $"{outerTypes}{genericArgs[genericArgsLowerBound].GetReadableName()}?";
             }
 
             // Remove the generic type count indicator (`n) from the type name
             // Avoid having to calculate the number of digits in the generic type count by assuming it's under 100
```
tjvezina, line 116: The second line of this comment can be removed now, as it's no longer using a shortcut

tjvezina (commented):
Thanks for fixing that! Just a couple of comments and a suggestion.

Mathematics/MathUtil.cs
mbaker3: Supposedly this is faster...well at least faster the smaller the number is....which seems like the most likely use case for this.

The SO post I linked in the doc comments the guy went through some tests. I haven't taken the time to validate the performance myself.
https://stackoverflow.com/a/51099524/640196

Reflection/TypeExtension.cs
mbaker3: Thanks, fixed!

Reflection/TypeExtension.cs
mbaker3: Good catch. Fixed!

tjvezina (dismissed):
The implementation of `GetDigitCount` makes me scowl, but it's certainly accurate.

jkeon (dismissed):

mbaker3:
Re-requesting review. After discussing I ended up moving all of the readability based utility methods into `ReadabilityExtension` and updating the tests as needed. Functionally this is the same but does introduce a breaking change in `anvil-unity-dots` which is resolved by this PR: https://github.com/decline-cookies/anvil-unity-dots/pull/96

Tests/Logging/ReadabilityExtensionTests.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            Assert.That(typeof(TestOuterGenericNestedClass<>.TestInnerGenericClass<>).GetReadableName(),
+                Is.EqualTo("ReadabilityExtensionTests.TestOuterGenericNestedClass<>.TestInnerGenericClass<>"));
+        }
+
+        [Test]
+        public static void GetDigitCount()
+        {
+            Assert.That(0.GetDigitCount(), Is.EqualTo(1));
+            Assert.That(0.GetDigitCount(true), Is.EqualTo(1));
```
jkeon, line 82: Neat, i didn't know you could call methods on a literal.

jkeon (approved):

Tests/Reflection/TypeExtensionTests.cs
```diff
@@ -1,32 +1,11 @@
-using NUnit.Framework;
-using Anvil.CSharp.Reflection;
+
 
 namespace Anvil.CSharp.Tests
 {
     public static class TypeExtensionTests
     {
```
tjvezina, line 6: Is this just an empty file now? Can it be removed?

mbaker3: #131 Incoming!

Tests/Logging/ReadabilityExtensionTests.cs
mbaker3: Yea, you do have to be careful about expressions like this though.
`-1.GetDigitCount(true)`. The `-` gets applied after the method call.

The fix is
`(-1).GetDigitCount(true)`

jkeon (approved):

PR #131, opened 2022-10-25
TypeExtensionTests - Implement

mbaker3:
Add tests for the remaining `TypeExtension` methods

`TypeExtension` does not have test coverage of `IsStatic()` and `CreateDefaultValue()`

`TypeExtension` has full test coverage

None

 - #130 

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Tests/Reflection/TypeExtensionTests.cs
+++ b/Tests/Reflection/TypeExtensionTests.cs
@@ -1,10 +1,36 @@
-
+using NUnit.Framework;
+using Anvil.CSharp.Reflection;
 
 namespace Anvil.CSharp.Tests
 {
     public static class TypeExtensionTests
     {
+        private class TestClass { }
+        private sealed class TestSealedClass { }
+        private static class TestStaticClass { }
+        private abstract class TestAbstractClass { }
+        private struct TestStruct { }
 
+
+        [Test]
+        public static void CreateDefaultValue()
+        {
+            Assert.That(typeof(TestClass).CreateDefaultValue(), Is.EqualTo(null));
+            Assert.That(typeof(TestSealedClass).CreateDefaultValue(), Is.EqualTo(null));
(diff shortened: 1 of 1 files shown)


jkeon (approved):

tjvezina:
Let's talk about test naming! There are many conventions, and no single standard, so it's down to personal preference and priorities.

In general the preference seems to be individual tests for each case, rather than one per function. This means that any UI that summarizes test results lets you see immediately which case failed, instead of only which function.

A common format is `MethodName_StateUnderTest_ExpectedResult()`[^1] (or `Given_When_Then`), ex. `IsStatic_WhenNonStaticClass_False()`, though some argue that including the name of the function being tested means manually updating tests when said function is renamed.

One suggestion is to add a separate test class (fixture) for each method[^2], grouping them together and removing the need to include the method name in every test name. This groups them nicely in test UIs as well.

Finally, a more recent standard goes for plain human readability[^3], for example `NonStaticClassIsNotStatic()` or `StaticClassTypeDefaultValueIsNull()` (note that names are definitive, not using "should"). This avoids including tested function names, leans into the idea that unit tests are also documentation, and may be more flexible and easier to write with NUnit's attributes (ex. multiple test cases).

Also, I arbitrarily created static classes for the first test fixtures, but it is recommended to create non-static classes[^4], if only because it's good practice, and to encourage test inheritance and reuse.

[^1]: [Naming standards for unit tests](https://osherove.com/blog/2005/4/3/naming-standards-for-unit-tests.html)
[^2]: [Structuring unit tests](https://haacked.com/archive/2012/01/02/structuring-unit-tests.aspx/)
[^3]: [You are naming your tests wrong!](https://enterprisecraftsmanship.com/posts/you-naming-tests-wrong/)
[^4]: [Static or non static test fixture in NUnit?](https://stackoverflow.com/questions/51836497/static-or-non-static-test-fixture-in-nunit#:~:text=NUnit%20supports%20static%20fixtures%20as,as%20well%20as%20maintain%20state.)

Issue #132, opened 2022-10-26
Refactor Unit Test Conventions


mbaker3:
Let's make the following changes:
 - [ ] Make the test classes `static`
 - [ ] Follow the naming convention here - https://haacked.com/archive/2012/01/02/structuring-unit-tests.aspx/
    - Basically the same as the other naming convention but I found these examples more clear.
 - [ ] Also follow the nested class per method outlined in the link above
    - This one feels like a lot of boiler plate but people seem to really like it and unit tests are already very boilerplate heavy as it is.

 - [ ] Apply updates to `anvil-unity-core`
 - [ ] Apply updates to `anvil-unity-dots`

The changes are out of scope for this PR but I'll create an issue that references this comment to get the work done.

_Originally posted by @mbaker3 in https://github.com/decline-cookies/anvil-csharp-core/issues/131#issuecomment-1292549667_

tjvezina:
Currently the test classes are already `static`, and there was an argument in the related PR for making them non-static to allow inheritance/re-use (though I'm not sure which I personally prefer yet) - just confirming that you didn't mean "Make the test classes non-`static`"?

mbaker3:
I did mean that. Fixed the description. Thanks

tjvezina:
Here's an example to confirm the proposed structure - I already like how it makes you put into words what you're testing, it certainly documents what's covered and helps come up with test cases. Is there anything you would change here?

---

`anvil-csharp-core/Tests/Data/Collections/Util/Array2DExtensionTests.cs` (Is the file name/folder structure good?)
```cs
namespace Anvil.CSharp.Tests
{
    public class Array2DExtensionTests
    {
        public class PopulateMethod
        {
            [Test]
            public void SetsAllValuesInArray()
            {
                // Assertions...
            }
        }

        public class AllMethod
        {
            [Test]
            public void ReturnsTrueWhenAllElementsMatch
            {
                // Assertions...
            }
            
            [Test]
            public void ReturnsFalseWhenSomeElementsMatch
            {
                // Assertions...
            }
            
            [Test]
            public void ReturnsFalseWhenNoElementsMatch
            {
                // Assertions...
            }
        }
    }
}
```

mbaker3:
That looks great. Only minor change I'd make to the naming is one method.

I think `AllValuesSetInArray` is more clear than `SetsAllValuesInArray`.
The new name focuses on what is being tested like the other method names.

tjvezina:
[...]

PR #137, opened 2022-12-05
TypeExtension.IsStatic name fix + Test Name Assertions

mbaker3:
- `TypeExtension.IsStatic` did not have the correct name casing
- Add test method name assertions to make sure test names are kept inline with method names if they get refactored.

- `TypeExtension.isStatic` is not the correct naming convention
- Refactored methods that have unit tests may not have their test names updated. 

- `TypeExtension.IsStatic`...promoted that `I`
- New assertions ensure that tests fail if a test's method name isn't updated to reflect the name of the method it is testing.

None

 - #136 

 - [x] Yes - Any calls to `isStatic` should be changed to `IsStatic`
 - [ ] No

### DIFF
--- a/Reflection/TypeExtension.cs
+++ b/Reflection/TypeExtension.cs
@@ -32,7 +32,7 @@ namespace Anvil.CSharp.Reflection
         /// This works because a static class is defined as abstract, sealed at the IL level
         /// Source: https://stackoverflow.com/a/1175901/640196
         /// </remarks>
-        public static bool isStatic(this Type type)
+        public static bool IsStatic(this Type type)
         {
             return type.IsAbstract && type.IsSealed;
         }
--- a/Tests/Data/Collections/Util/Array2DExtensionTests.cs
+++ b/Tests/Data/Collections/Util/Array2DExtensionTests.cs
@@ -9,6 +9,8 @@ namespace Anvil.CSharp.Tests
         [Test, Order(0)]
         public static void PopulateTest()
         {
+            Assert.That(nameof(PopulateTest), Does.StartWith(nameof(Array2DExtension.Populate)));
+
             const int SIZE = 5;
 
             int[,] array = new int[SIZE, SIZE];
@@ -26,6 +28,8 @@ namespace Anvil.CSharp.Tests
         [Test]
         public static void GetLengthTest()
         {
+            Assert.That(nameof(GetLengthTest), Does.StartWith(nameof(Array2DExtension.GetLength)));
+
             Assert.That(new int[1, 1].GetLength(), Is.EqualTo((1, 1)));
             Assert.That(new int[2, 3].GetLength(), Is.EqualTo((2, 3)));
             Assert.That(new int[24, 42].GetLength(), Is.EqualTo((24, 42)));
@@ -34,6 +38,8 @@ namespace Anvil.CSharp.Tests
         [Test]
         public static void AnyAllTest()
         {
+            Assert.That(nameof(AnyAllTest), Does.Contain(nameof(Array2DExtension.All)).And.Contain(nameof(Array2DExtension.Any)));
+
             bool[,] array = new bool[5, 5];
             array.Populate((x, y) => true);
 
@@ -53,6 +59,8 @@ namespace Anvil.CSharp.Tests
         [Test]
         public static void ForEachTest()
         {
+            Assert.That(nameof(ForEachTest), Does.StartWith(nameof(Array2DExtension.ForEach)));
+
             const int SIZE = 5;
 
             int[,] array = new int[SIZE, SIZE];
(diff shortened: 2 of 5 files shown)


tjvezina:
I'm not a fan of requiring all tests to start with an assert for the test name, and the format is going to change with #132 anyway. I see the benefit of automatic refactor handling, but I'm not sure the overhead is worth it - if you rename a function that's used in a unit test, you'll have to commit the changes to said test anyway, and will hopefully know (or the PR reviewer will notice) that the test name must be updated too.

At the very least, should it be `Does.Equal($"{nameof(SomeFunction)}Test")` instead of `StartsWith`, in case your refactoring involves simply shortening a function name? (Ex. `GetXValue` -> `GetX` would still match `GetXValueTest`)

mbaker3:
> ...the format is going to change with https://github.com/decline-cookies/anvil-csharp-core/issues/132

Ah right I forgot about that issue. The rules can be adjusted as part of that change but I think it's still possible to assert against the relevant names (maybe even easier?).

>  I see the benefit of automatic refactor handling, but I'm not sure the overhead is worth it - if you rename a function that's used in a unit test, you'll have to commit the changes to said test anyway, and will hopefully know (or the PR reviewer will notice) that the test name must be updated too.

I think the fewer things we have to remember as reviewers the better. It's so easy to miss something like that when looking at diffs. It's low effort to implement and provided me enough value during development to want to apply this convention to all tests.

@jkeon want to break a tie?

> At the very least, should it be Does.Equal($"{nameof(SomeFunction)}Test") instead of StartsWith, in case your refactoring involves simply shortening a function name? (Ex. GetXValue -> GetX would still match GetXValueTest)

The check is never going to be perfect although the test refactoring work may make this a more reliable check in the future. I can include the `...Test` portion to make it a better check for now.

jkeon:
> want to break a tie?

This is for whether the Test function name matches the function it's testing? The extra boiler plate is a little annoying, but I can see the benefit for keeping things nice and organized. 

It's not really part of the test though so I wonder if it would be better to have it as an Attribute? 

```csharp
[Test, Order(0)]
//Probably needs a better name for the attribute
[AnvilTestNamingConsistency(nameof(Array2DExtension.Populate)]
 public static void PopulateTest()
{
    etc
```

Code in the `PopulateTest` is solely for the test. As part of the `Setup` phase, you run a pass to gather all the methods with this attribute and then compare them to make sure they match the format?

I realize this is more work and could be a separate issue but I think the spirit is correct?



I would agree that we would want to check equality with the method names rather than StartsWith since we should know exactly what the method name should be.

mbaker3:
> This is for whether the Test function name matches the function it's testing? 

Whether we should have the assert at all I think was the main challenge. It seems like everyone identifies at least SOME benefit and then there's some question around whether it's worth it and how we go about it.

Given the discussion I'll adjust the name assertions to be a little more resilient to edge case name changes (Ex: `GetXValue` -> `GetX`).

If someone wants to think through a different approach (attributes, etc..) to propose in an issue they can. For now it's a low effort, best chance approach.

---

> It's not really part of the test though so I wonder if it would be better to have it as an Attribute?

I don't know. I'm not really bothered by the extra test. I agree that it is a bit of a tangent from the goal of the test method but I really can't see myself ever thinking it's worth the time to build out this feature.

> I would agree that we would want to check equality with the method names rather than StartsWith since we should know exactly what the method name should be.

This isn't necessarily true. There are unit tests that group the testing of multiple methods together. It's not the "right" way to test but sometimes it's just not practical.
Example: https://github.com/decline-cookies/anvil-unity-dots/blob/765b267e04e427b069e400fe7c3381b194356054/Scripts/Editor/Tests/Mathematics/MathExtensionTests.cs#L26

mbaker3:
Test name assertions adjusted and I did catch one test that wasn't correctly named.
In most cases it's not an exact `Is.EqualTo` check.

I've made exceptions and kept the `StartWith` but added a `"Test"` suffix to the name for situations where there are many tests for one util method.

The upside to this updated approach is that the test is more resilient against name refactoring.
The downside is that there's some manual work required if you want to change the method name for a test. That's not a big deal though since you're already be in there making edits.

jkeon (approved):

tjvezina (approved):
> There are unit tests that group the testing of multiple methods together.

I debated bringing this up - I agree that unit tests typically map to a single function but not always, yet that also goes against the planned reorganizing in #132, which explicitly groups tests by function.

For now at least, I'm fine with these changes, perhaps only because I hope to find an alternative whenever I get around to #132. I feel like as-is, it will be too easy in the future to add some unit tests and forget to include this `Assert` for the function name itself, but we'll see.

mbaker3:
> I debated bringing this up - I agree that unit tests typically map to a single function but not always, yet that also goes against the planned reorganizing in https://github.com/decline-cookies/anvil-csharp-core/issues/132, which explicitly groups tests by function.

Yea, I mean I could copy/paste the same tests into two different methods but that doesn't feel like a great solution. It's tough to test some of these methods in a vacuum. 

> I feel like as-is, it will be too easy in the future to add some unit tests and forget to include this Assert for the function name itself, but we'll see.

Agreed, a more automated approach would be best.

PR #141, opened 2023-01-13
Implement `DeferredWorkPump`

mbaker3:
Introduce a mechanism to defer work to a pre-defined, periodic, update phase.

#### Tag Alongs
 - Add `UpdateHandle.IsUpdating` to indicate whether the specific update handle instance is being updated.
    - This is in contrast to `UpdateHandle.IsSourceUpdating` which indicates whether the update source is currently updating for all `UpdateHandle`s listening to the same source.

Developers must devise their own approach to deferring and accumulating work to a periodic update phase. This usually means defining an `UpdateHandle`, `Queue` and optionally tracking whether an update is scheduled.

`DeferredWorkPump` is configured to update based on a user defined `AbstractUpdateSource`. Work may then be scheudled and accumulated against the `DeferredWorkPump` to be automatically executed at the desired update phase.

 - None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/DelayedExecution/DeferredWorkPump.cs
+++ b/DelayedExecution/DeferredWorkPump.cs
@@ -33,7 +33,7 @@ namespace Anvil.CSharp.DelayedExecution
         /// </param>
         public DeferredWorkPump(bool willEagerExecuteWork = false)
         {
-            m_ExecutePendingWorkStrategy = willEagerExecuteWork ? ExecutePendingWork_Eager : ExecutePendingWork;
+            m_ExecutePendingWorkStrategy = willEagerExecuteWork ? (Action)ExecutePendingWork_Eager : ExecutePendingWork;
 
             m_Update = UpdateHandle.Create<T>();
             m_PendingWork = new Queue<Action>();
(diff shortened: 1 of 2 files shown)


DelayedExecution/DeferredWorkPump.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            m_ExecutePendingWorkStrategy = willEagerExecuteWork ? (Action)ExecutePendingWork_Eager : ExecutePendingWork;
+
+            m_Update = UpdateHandle.Create<T>();
+            m_PendingWork = new Queue<Action>();
+        }
+
+        protected override void DisposeSelf()
+        {
+            m_Update.Dispose();
```
jkeon, line 44: Should we not clear `m_PendingWork`?

jkeon (approved):
Looks good, one additional clean up question

DelayedExecution/DeferredWorkPump.cs
mbaker3: Why waste the cycles on it? If this object is getting dereferenced it's not going to keep the actions referenced alive.

PR #145, opened 2023-02-09
Add DeferredDataPump

mbaker3:
Made the defer logic in `DeferredWorkPump` more generic so that any data (including delegates) could be accumulated and processed on an interval.

There is no convenient way to defer data to be processed at a specific interval.

`DeferredDataPump` can accumulate arbitrary data and process it with the provided callback providing an enumerator to iterate over the accumulated data.

The logic was moved from `DeferredWorkPump` which now extends `DeferredDataPump` to handle the bulk of its logic.

 - None

 - None

 - [x] Yes - The `ScheduleWork` method was renamed to `Schedule`
 - [ ] No


jkeon (approved):
Looks good!

PR #155, opened 2024-07-12
AbstractContentManager - Add IsShowing, Groups, GetAllShowingOfType

mbaker3:
- Allow callers to check if a specific view controller is showing in a group and, optionally, receive the instance via out parameter
 - Allow callers to get all showing of a specific view type
 - Allow callers to iterate groups

@tjvezina is this what you were wanting?
@jkeon, @tjvezina if this is an API we're happy with I'll take the time to clean it up and add `///` docs.

 - Callers have to know all content groups and check them individually to see if a specific piece of content is showing.
 - Callers have no way of iterating all groups in a content manager

#### IsShowing
Callers can call `AbstractContentManager.IsShowing<T>()` to figure out if a view is showing. Overloads and parameters of the method support.
 - Only looking for the exact type vs any subclass of the target type
 - Looking for a specific controller instance
 - Receiving the found content instance

#### GetAllShowingOfType
This is the plural version of `IsShowing` that returns all showing content of a given type. Includes the exact type vs subclass option of `IsShowing`

#### Groups Getter
Callers can get an `IReadonlyCollection<T>` of groups

 - None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Content/AbstractContentManager.cs
+++ b/Content/AbstractContentManager.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using Anvil.CSharp.Core;
 
 namespace Anvil.CSharp.Content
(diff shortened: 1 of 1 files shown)


tjvezina (dismissed):
This is great, thanks!

jkeon (approved):
Agreed, looks good!

mbaker3:
Ready for review.
 - Docs added
 - Removed generic parameter from the instance specific `IsShowing` method since it isn't required.

jkeon (approved):
Looks good!

PR #156, opened 2024-07-12
AbstractContentManager - Cleanup

mbaker3:
Make use of `Dictionary<,>.TryGetValue` in `AbstractContentManager`

There are a bunch of instances where we're checking if a key is contained in the group lookup and then fetching it manually.

Leveraging `Dictionary<,>.TryGetValue` accomplishes the same quicker and faster.

 - None

 - None

 - [ ] Yes 
 - [x] No


jkeon (approved):
Good change!

PR #157, opened 2024-08-29
Added `IndexOf` and `FindIndex` extensions for generic `IEnumerable<T>`

tjvezina:
C#'s enumerable classes are fairly inconsistent, for example if you use `List<T>.AsReadOnly()` it returns a `ReadOnlyCollection<T>` which is technically unordered. If you instead implicitly cast your list to `IReadOnlyList<T>` it allows indexing, but no longer implements `IndexOf`. Adding these extension methods just lends greater flexibility to working with enumerables, without having to convert to `List<T>` or `Array` to access certain functionality.

 - None

 - [ ] Yes
 - [X] No

### DIFF
--- a/Collections/EnumerableExtension.cs
+++ b/Collections/EnumerableExtension.cs
@@ -1,4 +1,6 @@
+using System;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Anvil.CSharp.Collections
 {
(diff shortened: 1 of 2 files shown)


jkeon (approved):

mbaker3:
Hmmm, I'm not sure about this. There's nothing about `IEnumerable` that guarantees consistent order.

I could see a backing collection changing in a refactor or a developer overlooking the assumptions that this extension method makes.

It would be safer if these extension methods were written for just `IReadOnlyList<T>`.

PR #162, opened 2025-12-05
ListExtension - Add Shuffle

mbaker3:
Shuffle elements of an IList using the Fisher-Yates algorithm

 - None

 - Implementations of `IList` can have their elements shuffled.

 - None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Collections/ListExtension.cs
+++ b/Collections/ListExtension.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 
 namespace Anvil.CSharp.Collections
 {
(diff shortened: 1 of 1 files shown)


Collections/ListExtension.cs
```diff
@@ -23,5 +24,20 @@ public static void AddUnique<T>(this IList<T> list, IEnumerable<T> items)
...
+        /// <param name="collection">The collection to shuffle the elements of</param>
+        /// <param name="randomProvider">Produces a random integer between arg1 (inclusive) and arg2 (exclusive)</param>
+        public static void Shuffle<T>(this IList<T> collection, Func<int, int, int> randomProvider)
+        {
+            int length = collection.Count;
+            for (int i = 0; i < length; i++)
+            {
+                int randomIndex =  randomProvider(0, length);
+                (collection[i], collection[randomIndex]) = (collection[randomIndex], collection[i]);
+            }
```
jkeon, line 40: I don't think this is a correct Fisher-Yates? 

https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle

I think the random Index needs to change on each iteration of the loop.

jkeon (changes requested):
Looks good, but implementation seems suspicious

Collections/ListExtension.cs
mbaker3: The random index is changing on each iteration? That's what `int randomIndex =  randomProvider(0, length);` does.

jkeon: Yeah but you're getting a random index from the entire set, 0 to length. Shouldn't it be i to length? The range should be changing each iteration according to the wikipedia article

mbaker3: Oh, yes. Good catch. Clearly I didn't understand the assignment.
Fixed

jkeon (approved):

